Issue #45, opened 2019-08-09
Publish dev version of component library when integration is merged


martypowell:
Ideally we can push a dev version of the component library to dev, for the design team to verify changes.

PR #46, opened 2019-08-09
Adjusted base font size given the design

martypowell:
This is mainly a test to verify that we can see our changes, but this also does add the appropriate baseline for font.


martypowell:
Addresses #45

Issue #38, opened 2019-08-08
Typography


martypowell:
## Component Name
N/A, but should include:
- `h1`
- `h2`
- `h3`
- `h4`-`h6` (these headings have  identical styling)
- `p`

## Please describe the requested component
Default functionality with styling see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements

**Please describe any required and optional properties for the component**

n/a

## Example or Mockup

Please see [design document.](https://xd.adobe.com/spec/e62d55ca-ca05-4c63-7049-04542c2f9ef2-901c/screen/140a31f7-1ebe-4590-8c2a-979b8fef6abc/Typography/)

## Acceptance Criteria

**Rule Based**:

- Must match visual styles defined in the design document.
- Must follow UX standards
- Semantically correct - Must follow [W3C](https://www.w3.org/) standards

## Additional context

- These will not require any react components, but will require a documentation file. Not sure how this will work. (-MP)
- These are based on the heading and text styles that will be most common on the site. A number of alternative and context-specific versions of headings and p tags will be required in future springs.

danfox01:
@martypowell what sort of documentation do you need for this? Style info from Sam?

martypowell:
Yea i think so, this one is kinda setup, but it should be logged.

danfox01:
@sdurphy can you please provide the details for this? Font faces, weights, sizes for body, headings etc.

danfox01:
@martypowell @sdurphy 

Need the following base fonts , I think:

- Heading 1
- Heading 2
- Heading 3
- Heading 4 (5 and 6 styled the same as 4)
- Paragraph
- Ordered list
- Unordered list
- Block Quote

PR #85, opened 2019-08-21
Typography

martypowell:
Addresses #38 

- Updates styles for typography
- Adds section component in order to display typography or a dark section


martypowell:
@danfox01 @sdurphy I pushed the example up knowing that it probably needs some formatting, but would like you guys to review first. Another issue that needs resolved is the margins/padding for all elements.

danfox01:
@sdurphy the Montserrat body copy seems too heavy. Currently set at 600. Should it be 500?

danfox01:
@martypowell I think we don't want the heading border to span the width of the column. Switching headings to `inline-block` might solve?

sdurphy:
Before approving, can we make the following changes:

- In the Paragraph section we need to change #333 to #292929.
- ~~Can we add 1px of padding to the left and right of the link for Link Focus? (Normal Link, and Normal Link - Sidebar)~~

sdurphy:
Ignore my second bullet about the Link Focus... I was hoping it wouldn't bump anything, but after testing it, it does.

martypowell:
> @martypowell I think we don't want the heading border to span the width of the column. Switching headings to `inline-block` might solve?

@danfox01 we are going to leave this as for this pr. A heading is a block element and I think we will want to keep it that way. The underline is something we will address as part of the inside pages once it comes up.

martypowell:
> Ignore my second bullet about the Link Focus... I was hoping it wouldn't bump anything, but after testing it, it does.

Ignored.

martypowell:
> Before approving, can we make the following changes:
> 
> * In the Paragraph section we need to change #333 to #292929.
> * ~Can we add 1px of padding to the left and right of the link for Link Focus? (Normal Link, and Normal Link - Sidebar)~

@sdurphy wording changed. ready for another look.

sdurphy:
@danfox01 @martypowell, it looks like the font-weight for the right column body copy is too heavy. Can that be brought down to 500 (medium)?

martypowell:
> @danfox01 @martypowell, it looks like the font-weight for the right column body copy is too heavy. Can that be brought down to 500 (medium)?

@sdurphy adjusted

danfox01:
@martypowell Kind of minor for now but we need to update the example text for `Normal Paragraph - Sidebar (Right column on desktop)`. It is just a copy/paste of the other example, but it should be updated to reflect the actual font name (Montserrat) and size (16px). @sdurphy can clarify.

martypowell:
> @martypowell Kind of minor for now but we need to update the example text for `Normal Paragraph - Sidebar (Right column on desktop)`. It is just a copy/paste of the other example, but it should be updated to reflect the actual font name (Montserrat) and size (16px). @sdurphy can clarify.

@danfox01 in the latest update.

danfox01:
Great. On my phone right now and can’t actually hit approve button. Feel free to push through.

Sent with OWA for iPhone

sdurphy (approved):

Issue #32, opened 2019-08-08
Textarea


martypowell:
**Please name the requested component**
TextArea

**Please describe the requested component**
Component should use default functionality based on https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textareaas a guideline

**Please describe any required and optional properties for the component**

- **Hint** - type - _string_ - only when absolutely necessary, adds text to help the user fill out a field, ex Password Requirements
- **Label** - type - _string_ - Label to describe the associated text input
- **Id** - type - string - Unique identifier used to associate the input with the label

**Example or Mockup**

Should look most likely the [text input](https://github.com/baltimorecounty/dotgov-components/issues/28)

**Acceptance Criteria**

**Rule Based**:

- Each text input requires a label
- Label when selected must focus on the input
- When hint is present, visible underneath the label
- Must follow UX standards
- Semantically correct - Must follow [W3C](https://www.w3.org/) standards
- Matches Mockup Styles

**Additional context**

The textarea to be equivalent to 5 lines of input text for minimum height.

martypowell:
@danfox01 once piece of information we probably should include is a `min-height`?

danfox01:
@martypowell, good point. I think we want the `textarea` to be equivalent to 5 lines of input text. So whether that is accomplished with something like `5em` or a fixed height if up to you, but the result should be the user being able to see 5 full lines of textbefore scrolling.

Issue #28, opened 2019-08-07
Text Input


martypowell:
**Please name the requested component**
TextInput

**Please describe the requested component**
Component should use default functionality based on https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input as a guideline

**Please describe any required and optional properties for the component**

- Hint - type - string - only when absolutely necessary, adds text to help the user fill out a field, ex Password Requirements
- Label - type - string - Label to describe the associated text input
- Id - type - string - Unique identifier used to associate the input with the label

**Example or Mockup**

_Default State_

![image](https://user-images.githubusercontent.com/1933400/62624954-42926e80-b8f2-11e9-8611-3064ef6046ab.png)

_Focused_

![image](https://user-images.githubusercontent.com/1933400/62625056-7372a380-b8f2-11e9-8b4c-fea0b3636160.png)

**Acceptance Criteria**

**Rule Based**:

- Each text input requires a label
- Label when selected must focus on the input
- When hint is present, visible underneath the label
- Must follow UX standards
- Semantically correct - Must follow [W3C](https://www.w3.org/) standards
- Matches Mockup Styles

**Additional context**

We will want to allow for this field to be validated, but that can be handled independently of this change.

sdurphy:
Just FYI all, the focus color should be #1076BC (Blue Vibrant), not gold.

PR #86, opened 2019-08-21
Textarea

martypowell:
Addresses #32 

- Added component
- Added styles

### DIFF
--- a/src/components/basic/form/TextArea.js
+++ b/src/components/basic/form/TextArea.js
@@ -1,0 +1,23 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import FormField from './FormField';
+
+const TextArea = (props) => {
+	const { id, label, hint, ...otherProps } = props;
+	return (
+		<FormField id={id} label={label} hint={hint}>
+			<textarea className="dg_form-field_textarea" id={id} {...otherProps} />
+		</FormField>
+	);
+};
+
+TextArea.propTypes = {
+	/** A unique identifier used to associate the textarea with the label */
+	id: PropTypes.string.isRequired,
+	/** Only when absolutely necessary, adds text to help the user fill out a textarea */
+	hint: PropTypes.string,
+	/** A label to describe the associated textarea */
+	label: PropTypes.string.isRequired
+};
+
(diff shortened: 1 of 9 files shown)


danfox01:
@martypowell looks good except there's a typo in the closing span of the non-React HTML snippet:

`<span class="dg_label-text">Describe your problem/span>`

martypowell:
> @martypowell looks good except there's a typo in the closing span of the non-React HTML snippet:
> 
> `<span class="dg_label-text">Describe your problem/span>`

good catch, will fix.

martypowell:
> > @martypowell looks good except there's a typo in the closing span of the non-React HTML snippet:
> > `<span class="dg_label-text">Describe your problem/span>`
> 
> good catch, will fix.

@danfox01 addressed.

danfox01:
@martypowell we don't want to force all caps for the label or hint. As we discussed this also affects the standard text input, #28.

It also looks like the input label text is too large, but we failed to document that anywhere so I will create a separate issue for that.

martypowell:
> @martypowell we don't want to force all caps for the label or hint. As we discussed this also affects the standard text input, #28.
> 
> It also looks like the input label text is too large, but we failed to document that anywhere so I will create a separate issue for that.

@danfox01 I am resolving now. Confirmed with Sam, label text is 18px.

martypowell:
> > @martypowell we don't want to force all caps for the label or hint. As we discussed this also affects the standard text input, #28.
> > It also looks like the input label text is too large, but we failed to document that anywhere so I will create a separate issue for that.
> 
> @danfox01 I am resolving now. Confirmed with Sam, label text is 18px.

@danfox01 @sdurphy please both review. Bugs that arose should be resolved

- Label and Hint styling
- Bad input focus

danfox01:
@martypowell I'm actually confused about where I can review these changes. The Now link above still shows the previous version.

martypowell:
> @martypowell I'm actually confused about where I can review these changes. The Now link above still shows the previous version.

@danfox01 thats because I never pushed up my changes, it should be deploying now.

sdurphy (approved):

Issue #39, opened 2019-08-08
Bulleted Lists


martypowell:
**Please name the requested component**
- OrderedList
- UnOrderedList

Each will contain a `ListItem`

**Please describe the requested component**
Default functionality based on 

- ul - https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ul
- ol - https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ol

**Please describe any required and optional properties for the component**

n/a

**Example or Mockup**

See "Content Area Lists" https://xd.adobe.com/spec/e62d55ca-ca05-4c63-7049-04542c2f9ef2-901c/screen/140a31f7-1ebe-4590-8c2a-979b8fef6abc/Typography

**Rule Based**:

- Ordered lists must support two levels: Numbers at the first level, and bullets at the second level.
- Unordered lists must support two levels: Solid bullets at the first level, and hollow bullets at the second level.
- Lists should inherit font styling (font face, line height, size, etc) from the surrounding body content.
- Must follow UX standards
- Semantically correct - Must follow [W3C](https://www.w3.org/) standards
- Matches Mock-up Styles

**Additional context**
n/a

danfox01:
Updated acceptance criteria and added link to design doc.

PR #87, opened 2019-08-23
Lists

martypowell:
Addresses #39 

- Adds component for `ordered` and `unordered` list
- Adds styles for lists
- Removed list reset logic


danfox01 (approved):
One note from @jdomasky's QA. The bullets in the build are larger than those in the comp. Confirmed with @sdurphy that the larger bullet size is correct. This is good to go.

Issue #27, opened 2019-04-25
Table


danfox01:
## Please name the requested component
Table

## Description
- This will be the default styling for all `<table>` elements.
- Table should be striped and allow for row highlighting on hover, behavior similar to [current tables on the .gov](https://www.baltimorecountymd.gov/News/holidays/index.html).

## Required and optional properties
n/a

## Example or Mockup

![FireShot Capture 120 - Dot-Gov-Components - xd adobe com](https://user-images.githubusercontent.com/13049389/63195655-353b5980-c041-11e9-845a-71a9cf8d48c8.png)

See the [design document.](https://xd.adobe.com/spec/e62d55ca-ca05-4c63-7049-04542c2f9ef2-901c/screen/4aa3de60-a9f6-47f1-a746-3ee6e5f19d68/Tables)

## Acceptance Criteria

**Rule Based:**

- Table design, colors, fonts, etc. ,must match design document.
- Table rows must have alternatively striped via CSS, using the colors in the design document.
- Table rows must have a hover state using light gold, as specified in design document.
- Must follow UX standards
- Semantically correct - Must follow [W3C](https://www.w3.org/) standards

## Additional context
In addition to the default table style, we may need a lightweight version (potentially less spacing within cells and a slightly smaller font size) that allows for more content to be shown at once, for data-heavy tables. This decision might need to be TBD based on the how the default table performs.

martypowell:
@danfox01 is there any case where column sorting wouldn't be available?

danfox01:
@martypowell i think it only makes sense in certain cases.  Specifically, small tables probably don't need it, and certain content would never need to be sorted. The example above is a good one -- sorting doesn't provide value and would probably look weird.

martypowell:
@danfox01 ok, so i think its fair to add the basic table initially until we hit one that needs the sorting and potentially filtering?

danfox01:
@martypowell Yep. For this sprint for sure, basic table. The sorting and filtering would prob be more of an advanced component. Updated description above to reflect this.

danfox01:
Larger table example from content area of mockup:
![table](https://user-images.githubusercontent.com/13049389/62872396-60dddd00-bceb-11e9-9726-91a3f9c349e4.jpg)

danfox01:
Added acceptance criteria.

PR #88, opened 2019-08-23
Table

theilman7:
#27 Pushing code for table component. This creates the table component for use on the new DG website.


src/components/basic/table/Table.md
```diff
@@ -0,0 +1,102 @@
...
+        <th>Crust</th>
+        <td>Thin</td>
+        <td>Deep Dish</td>
+      </tr>
+    </tbody>
+  </table>
+</div>
+```
+
+### Table Dark
```
martypowell, line 71: can you please render an example for dark, so it's visible in the styleguide?

src/components/basic/table/TableBody.js
```diff
@@ -0,0 +1,8 @@
+import React from "react";
+import ReactDOM from "react-dom";
```
martypowell: You can remove any of these references in these files, we aren't using them.

src/components/basic/table/TableCell.js
```diff
@@ -0,0 +1,8 @@
+import React from "react";
+import ReactDOM from "react-dom";
```
martypowell: You can remove any of these references in these files, we aren't using them.

src/components/basic/table/TableHead.js
```diff
@@ -0,0 +1,8 @@
+import React from "react";
+import ReactDOM from "react-dom";
```
martypowell: You can remove any of these references in these files, we aren't using them.

src/components/basic/table/TableHeadCell.js
```diff
@@ -0,0 +1,8 @@
+import React from "react";
+import ReactDOM from "react-dom";
```
martypowell: You can remove any of these references in these files, we aren't using them.

src/styles/partials/components/_table.scss
```diff
@@ -0,0 +1,88 @@
+/* BASIC TABLE STYLES */
+.table {
+  border: 1px solid $gray-light;
+  border-collapse: collapse;
+  color: #4c4c4c;
```
martypowell: this color needs to come from the variables file. If it isn't in the color list, we need to talk to sam.

src/styles/partials/components/_table.scss
```diff
@@ -0,0 +1,88 @@
...
+  font-family: $default-heading-font;
+  font-size: 16px;
```
martypowell: this needs to come from teh font variables file.

src/styles/partials/components/_table.scss
```diff
@@ -0,0 +1,88 @@
...
+  text-align: left;
+  width: 100%;
+  td {
+    border-bottom: 1px solid $gray-light;
```
martypowell: we should add `1px` to the borders variable file as `$border-width-smallest`

src/styles/partials/components/_table.scss
```diff
@@ -0,0 +1,88 @@
...
+    padding: 15px;
+  }
+  thead {
+    th {
+      background-color: $white;
+      border-bottom: 5px solid $blue;
+      border-left: none;
+      color: $black;
+      font-family: $default-heading-font;
+      font-size: 18px;
```
martypowell: should come from the font variables.

src/styles/partials/components/_table.scss
```diff
@@ -0,0 +1,88 @@
...
+      font-weight: bold;
+      padding: 15px;
```
martypowell: this should come from the spacing variables

src/styles/partials/components/_table.scss
```diff
@@ -0,0 +1,88 @@
...
+    }
+  }
+  th {
+    border-bottom: 1px solid $gray-light;
+    border-left: 5px solid $blue;
+    color: $black;
+    padding-left: 20px;
```
martypowell: this should come from the spacing variables file

src/styles/partials/components/_table.scss
```diff
@@ -0,0 +1,88 @@
...
+      color: $black;
+      font-family: $default-heading-font;
+      font-size: 18px;
+      font-weight: bold;
+      padding: 15px;
+    }
+  }
+  th {
+    border-bottom: 1px solid $gray-light;
+    border-left: 5px solid $blue;
```
martypowell: thse border values should come from the borders variables file

src/styles/partials/components/_table.scss
```diff
@@ -0,0 +1,88 @@
...
+      }
+      &:hover {
+        background-color: $gold-lightest;
+      }
+    }
+  }
+}
+
+/* BLUE TABLE STYLES */
+.dg_section.dark {
```
martypowell: this logic should move to the sections file and should not move over any properties are duplicated in the default table.

martypowell (changes requested):
Please take a look at the comments, most of the properties should use variables, most of which exist, if they don't let me know.

danfox01:
FYI, I will hold my review until after the requested code changes are made.

src/components/basic/table/TableBody.js
theilman7: Removed from components locally. Will include a note when I commit the update version.

src/styles/partials/components/_table.scss
theilman7: All fugitive hex colors have been replaced with their corresponding variables. Will include update in next commit.

src/styles/partials/components/_table.scss
theilman7: Didn't realize there was a $border-width-small variable in that file that is set to 1px. I can use that unless you would like it renamed to $border-width-smallest.

src/styles/partials/components/_table.scss
theilman7: Updated the font-size: 18px; with the $small-heading-font-size variable. Will include this in my next commit.

src/styles/partials/components/_table.scss
theilman7: Replaced with $padding-large. Will update in my next commit.

src/styles/partials/components/_table.scss
martypowell: no problem, yes please use.

src/styles/partials/components/_table.scss
martypowell: @theilman7 I actually thought about this a little this weekend. And im torn of whether the dark sections css should all go in teh sections file or if they should go in the component file. Do you have any thoughts?

src/styles/partials/components/_table.scss
theilman7: All fugitive border values have been updated with their corresponding variables.

src/styles/partials/components/_table.scss
theilman7: @martypowell I'm leaning towards the component file. I feel like keeping the dark section styes in the component file will help tidy the structure, since the dark section version of the component, is essentially just a different variation of the base component in relation to the overall taxonomy. Looking ahead, it seems more intuitive that the dark section version of the component would be listed in the main component file, than someone fresh, coming in and heading straight for the component file, not seeing the dark section version and having to then find the section file. In addition, I feel like adding every dark section component style to the section file would create quite a large and muddied file after awhile.

martypowell: i think this is a good point. Here is the argument against that approach

- Say we add a `.light` section. In that case, we would have to go into each component (and know that component has styles for that section, which wouldn't be horrible, becasue we could look for `.dg_section.dark`)
- If they are separated into a section file, then we can copy and paste the existing dark section and make the changes as needed
- Say we need to remove the dark section, then we have to touch all these different file parts instead of one location.

Do these points impact your thoughts at all?

danfox01:
@theilman7 @martypowell Give me a heads up when all the code issues have been addressed and I will do my review.

src/components/basic/table/TableBody.js
```diff
@@ -0,0 +1,7 @@
+import React from "react";
+
+const TableBody = props => {
+	return <tbody>{props.children}</tbody>;
```
martypowell: in this case this can be

```js
//(param1, param2, …, paramN) => expression
// equivalent to: => { return expression; }

const TableBody = props => <tbody>{props.children}</tbody>;
```

See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#Basic_syntax

martypowell:
@danfox01 this is ready for review, there will still be a couple minor changes, but nothing noticeable.

src/styles/partials/components/_table.scss
martypowell: just so its documented we decided to go {component}-{theme} to for a pattern. It documented in the main sass file and we will evaluate this pattern as the project gets bigger.

@theilman7

src/styles/partials/components/_table-dark.scss
```diff
@@ -0,0 +1,46 @@
+.dg_section {
+	&.dark {
+		.table {
+			background-color: $pale-blue;
+			border: $border-width-small solid $gray-light;
+			border-collapse: collapse;
```
martypowell: if you take a look at `src/styles/partials/components/_table.scss` some of these styles are repeated. In this case, we don't need to repeat something like `border-collapse` because it is already include for the default table. the only styles we will want to include in this particular set of rules are the ones that make it look ok on a dark background, mostly colors.

danfox01:
@theilman7 @martypowell Looks like the example in the HTML snippet doesn't match the code on the page. Otherwise, looks great.

src/styles/partials/components/_table-dark.scss
theilman7: I will work on trimming this down and will commit after.

theilman7: Revisions have been committed.

src/components/basic/table/TableBody.js
theilman7: Revisions have been committed.

martypowell (approved):

martypowell (dismissed):

martypowell:
@danfox01 this is good code wise, just needs approval from a design perspective

danfox01:
@theilman7, the markup example looks great now. One more small change (sorry): the weight for the table heads should actually be `600`. Currently, they are set to `bold`, which is equal to 700.

In the future we probably want to use numeric weight values instead of the strings, but @martypowell can comment on best practice.

FYI @sdurphy

martypowell:
@danfox01 @theilman7 made the changes, please look again.

danfox01:
Looks good to me.

martypowell (approved):
@theilman7 I wasn't able to remove fieldset, but it shouldn't cause any issues, so allowing this through.

Issue #37, opened 2019-08-08
Button


martypowell:
**Please name the requested component**
Button

**Please describe the requested component**
Use the default behavior based on https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button

**Please describe any required and optional properties for the component**
Only styles are applied right now.

**Example or Mockup**

https://xd.adobe.com/spec/6c7ed8ba-6428-418d-464e-d31fb3608c6b-1a4f/

**Acceptance Criteria**

**Rule Based**:

- Must handle the followings states
  - default
  - focused
  - hovered
  - disabled
- Must follow UX standards
- Semantically correct - Must follow [W3C](https://www.w3.org/) standards
- Matches Mockup Styles

**Additional context**

This might be a little fluid, and developers should work with designers on this component closely.

martypowell:
@danfox01 in baltcogo we have the option to include an icon in the button, is that something we want to include in this system?

martypowell:
> @danfox01 in baltcogo we have the option to include an icon in the button, is that something we want to include in this system?

this is different than icons buttons, in this case an example would be.

default state: Submit Form
sumbitting state: loading icon plus submitting form

something along those lines.

martypowell:
@danfox01 not sure where this gets documented, but wanted to call out that a button must be included in a snippet, form, or an external app. Limitation of site executive.

Whether this is important or not, I don't know.

martypowell:
@sdurphy well need to be sure to handle all of the states.  https://cloudfour.com/thinks/designing-button-states/

I believe we want to cover the following?

- default
- focused
- hovered
- disabled

fyi @danfox01

danfox01:
Wow @martypowell... getting caught up. We can mock up all button types and all states, and include above. @sdurphy, I'll hit you up this thread to keep it from blowing up.

martypowell:
@danfox01 @sdurphy referencing the inspiration for this

https://designsystem.digital.gov/components/button/

PR #89, opened 2019-08-26
Buttons

martypowell:
Addresses #37 

- Primary Buttons
- Secondary Buttons
- Link Buttons
- Icon Links

Adds animation to hover for buttons and links. Added web fonts to display font awesome.

Does not add navigation buttons, that will be handled as part of the navigation build


martypowell:
@martypowell please add examples for the button usage to the docs.

src/styles/partials/components/_button.scss
```diff
@@ -0,0 +1,92 @@
...
+	font-weight: $font-bold;
+	padding: $default-padding;
+	text-transform: uppercase;
+
+	i {
+		display: inline;
+		margin-right: $margin-small;
+	}
+
+	&:active,
```
martypowell: these can be simlified a bit.

martypowell: simplified as much as possible

src/components/basic/IconLink.js
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+
+const IconLink = (props) => {
+	const sizes = {
```
martypowell: we should probably just control this through a class, in order to simplify this. Be sure to update the docs.

martypowell: this is done.

src/styles/partials/components/_link.scss
```diff
@@ -0,0 +1,79 @@
...
+	&.disabled {
+		color: $gray-light;
+
+		i {
+			color: $white;
+			background: $gray-light;
+		}
+	}
+
+	&.dg_icon-link--small {
```
martypowell, line 69: this looks like a good case for a mixin

danfox01 (changes requested):
@martypowell looks great, except for a few things:

- Hover state of `dg_button-link` in a dark section is incorrect. The background color should not change on hover, instead the link color should change.
- Font weight of `dg_button-link` should be 600 (semibold) not 700 (bold)
- I noticed we're not currently pulling the italic version of Crimson. We need to add this for true italics.

FYI @sdurphy

src/styles/partials/components/_link.scss
martypowell: done in latest changes.

martypowell:
> @martypowell looks great, except for a few things:
> 
> * Hover state of `dg_button-link` in a dark section is incorrect. The background color should not change on hover, instead the link color should change.
> * Font weight of `dg_button-link` should be 600 (semibold) not 700 (bold)
> * I noticed we're not currently pulling the italic version of Crimson. We need to add this for true italics.
> 
> FYI @sdurphy

Hey @sdurphy @danfox01 did these get addressed?

danfox01:
Per testing by @jdomasky, a few inconsistencies and variation from the comps:

- For `dg_icon-link` in a dark section, the hover color should be `blue-lightest` not `blue-vibrant`.
- Most of the active state examples also respond to hover. The exception seems to be `dg_icon-link` and `dg_icon-link--small`, which do not show a hover state. Seems weird to me that an example of an active state would have a hover, so I'm not sure which approach is correct. For reference, the [active button examples in the USDS style guid](https://designsystem.digital.gov/components/button/)e do not respond to hover. Regardless, our examples should be consistent.

It's also worth noting that the buttons sizes may need to be tweaked for mobile in future iterations. This might be related to responsive typography.

src/styles/partials/components/_link.scss
```diff
@@ -0,0 +1,80 @@
...
+	text-align: center;
+	text-decoration: none;
+	text-transform: uppercase;
+
+	i {
+		background: $gold;
+		color: $black;
+		border-radius: 50%;
+		display: inline-block;
+		line-height: 100px;
```
theilman7, line 25: This is more of a question. I'm wondering if we should include values for line-height and other CSS typographic focused variables in the _fonts.scss file. IE. $line-height-lg, $line-height-sm etc.

src/styles/partials/base/_button-reset.scss
```diff
@@ -0,0 +1,26 @@
...
+input[type=submit],
+input[type=reset] {
+	border: none;
+	margin: 0;
+	padding: 0;
+	width: auto;
+	overflow: visible;
+
+	background: transparent;
+
```
theilman7, line 11: For consistency and for cleaning up the markup, I would add the background property before border so we can maintain the alphabetic order of css properties.

src/styles/partials/components/_button.scss
```diff
@@ -0,0 +1,105 @@
...
+		border-color: $vibrant-blue;
+	}
+
+	&:disabled,
+	&.disabled {
+		background: $gray-light;
+		border-color: $gray-light;
+	}
+}
+
```
theilman7, line 50: For consistency, move the border-color property below the background property.

src/styles/partials/components/_label.scss
```diff
@@ -5,7 +5,7 @@
```
theilman7, line 5: These properties should be in the following order for consistency: font-size, font-weight, margin-top.

src/styles/partials/base/_button-reset.scss
martypowell: i copied this from somewhere, else, but I'm good wtih this comment.

src/styles/partials/components/_button.scss
martypowell: good catch, done.

src/styles/partials/components/_label.scss
martypowell: adjusted.

src/styles/partials/components/_link.scss
martypowell: I don't have a strong knowledge of how to effectively use line height, it's always more have been a case by case basis for me personally. If i you have a good handle on this, then I'd defer to you. But I'm always in favor of creating some more rules around this.

src/styles/partials/base/_button-reset.scss
martypowell: adjusted any no commented proprties

martypowell:
> Per testing by @jdomasky, a few inconsistencies and variation from the comps:
> 
> * For `dg_icon-link` in a dark section, the hover color should be `blue-lightest` not `blue-vibrant`.
> * Most of the active state examples also respond to hover. The exception seems to be `dg_icon-link` and `dg_icon-link--small`, which do not show a hover state. Seems weird to me that an example of an active state would have a hover, so I'm not sure which approach is correct. For reference, the [active button examples in the USDS style guid](https://designsystem.digital.gov/components/button/)e do not respond to hover. Regardless, our examples should be consistent.
> 
> It's also worth noting that the buttons sizes may need to be tweaked for mobile in future iterations. This might be related to responsive typography.

Looking now. I also think i got the clasess wrong, looking now. Since doing this pr I realized I should have done one button set a time :disappointed:

martypowell:
> > Per testing by @jdomasky, a few inconsistencies and variation from the comps:
> > 
> > * For `dg_icon-link` in a dark section, the hover color should be `blue-lightest` not `blue-vibrant`.
> > * Most of the active state examples also respond to hover. The exception seems to be `dg_icon-link` and `dg_icon-link--small`, which do not show a hover state. Seems weird to me that an example of an active state would have a hover, so I'm not sure which approach is correct. For reference, the [active button examples in the USDS style guid](https://designsystem.digital.gov/components/button/)e do not respond to hover. Regardless, our examples should be consistent.
> > 
> > It's also worth noting that the buttons sizes may need to be tweaked for mobile in future iterations. This might be related to responsive typography.
> 
> Looking now. I also think i got the classess wrong, looking now. Since doing this pr I realized I should have done one button set a time :/

@danfox01 I think all issues have been addressed, please review.

danfox01:
Per conversation with @martypowell final review of this issue is on hold until later this week. When we resume, here are retest results from @jdomasky:

> 1. The hover class was removed from all of the Active buttons except for one: Text Button in a Dark Section.
> 2. In the Decorative Links in a Dark Section (both Large and Small), the Disabled buttons have a hover class configured. I believe this is a new bug. I don't remember observing hover on these buttons when we tested yesterday.

martypowell:
> Per conversation with @martypowell final review of this issue is on hold until later this week. When we resume, here are retest results from @jdomasky:
> 
> > 1. The hover class was removed from all of the Active buttons except for one: Text Button in a Dark Section.
> > 2. In the Decorative Links in a Dark Section (both Large and Small), the Disabled buttons have a hover class configured. I believe this is a new bug. I don't remember observing hover on these buttons when we tested yesterday.

Both valid bugs and addressed @jdomasky ready for another look.

fyi @danfox01

jdomasky:
@martypowell @danfox1 Taking another look.

jdomasky:
@martypowell @danfox1 The issues in the retest results from this morning have been resolved. However, there's still an issue with this component:

- In the Secondary Button set and the Text Button in a Dark Section, the Disabled buttons have a hover class configured. I'm not sure if this was overlooked during retesting or if this issue was just introduced.
**Recommendation:** Remove the hover class from these two Disabled buttons.

martypowell:
> @martypowell @danfox1 The issues in the retest results from this morning have been resolved. However, there's still an issue with this component:
> 
> * In the Secondary Button set and the Text Button in a Dark Section, the Disabled buttons have a hover class configured. I'm not sure if this was overlooked during retesting or if this issue was just introduced.
>   **Recommendation:** Remove the hover class from these two Disabled buttons.

@jdomasky this is a bug, and was just overlooked, fixing now.

src/styles/partials/components/_button.scss
martypowell: @theilman7 ready for another look

src/styles/partials/components/_label.scss
martypowell: @theilman7 ready for another look

src/styles/partials/components/_link.scss
martypowell: > > @martypowell @danfox1 The issues in the retest results from this morning have been resolved. However, there's still an issue with this component:
> > 
> > * In the Secondary Button set and the Text Button in a Dark Section, the Disabled buttons have a hover class configured. I'm not sure if this was overlooked during retesting or if this issue was just introduced.
> >   **Recommendation:** Remove the hover class from these two Disabled buttons.
> 
> @jdomasky this is a bug, and was just overlooked, fixing now.

@jdomasky this is ready for another look.

danfox01:
@martypowell , this is good to merge.

src/styles/partials/components/_link.scss
jdomasky: @martypowell The buttons are good to go. All testing issues have been resolved.

PR #90, opened 2019-08-27
created select component#35

sgrg1:
create select component with .md file and test.js . all passed


src/styles/partials/components/_input.scss
```diff
@@ -1,6 +1,6 @@
 $input-border-width: 2px;
 
-input,
+input,select,
```
martypowell: select should be on a new line.

```css
#selector1,
.selector2,
.test < test
... 
```

src/components/basic/form/Select.test.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import Select from './Select';
+
+const options =[{id:1,name:'published'},{id:2,name:'updated'},{id:3,name:'views'},{id:4,name:'comments'}];
```
martypowell: please format this list.

```js
const options = [
    {
        id: 1,
...
```

martypowell (commented):
@sgrg1 this is a little off, but we need to provide a mockup, getting one from @sdurphy

src/components/basic/form/Select.md
```diff
@@ -0,0 +1,26 @@
...
+  options={options}
+/>
+```
+
+If you aren't using react, you'll need to use the following html:
+
+```html
+<div class="dg_form-field">
+  <label for="select-items" class="dg_label">
+    <span class="dg_label-text">Select from the list</span>
```
martypowell: For the example this should be clear to a non developer. So we want this to describe what is being selected below.

I'm not sure what the list below contains. Something that might be more clear would be

Label: Favorite Colors
Options: Blue, Red, Yellow, Gray, Black

src/styles/partials/components/_input.scss
sgrg1: complete

src/components/basic/form/Select.test.js
sgrg1: complete

src/components/basic/form/Select.md
sgrg1: complete

martypowell:
@sdurphy @danfox01 I did some research on implementing the arrow based on the mockup and recommend against it. We can make this visually work with a container around the select, but a portion of the select is then not selectable see the red:

image

This would require JavaScript to get involved, which I don't think we are at that point that we need that right now.

Please let @sgrg1 and I know if we should proceed.

src/components/basic/form/Select.md
```diff
@@ -0,0 +1,27 @@
...
+
+If you aren't using react, you'll need to use the following html:
+
+```html
+<div class="dg_form-field">
+  <label for="select-items" class="dg_label">
+    <span class="dg_label-text">Favorite Colors</span>
+  </label>
+  <select id="select-items" className="dg_form-field-select">
+      <option value="">- Select a Favorite Color -</option>
```
martypowell: formatting if off here.

src/components/basic/form/Select.test.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import Select from './Select';
+
+const options =[
+	{
```
martypowell, line 6: formatting is off, when you get to this please let me knwo, and we can try and configure your editor to do this automatically.

src/styles/partials/components/_input.scss
```diff
@@ -1,6 +1,7 @@
 $input-border-width: 2px;
```
martypowell: this value should come from the _variables < _border file.

martypowell: so you can remove it from line one, and use the variable from borders here:

```sass
border: $input-border-width solid $black;
```

danfox01:
Functionally, this looks great. @martypowell, per your comment above, we will look into a custom drop down icon as a backlog item -- not for this sprint.

One change since Sam finalized the design doc: The font for the dropdown text should be Montserrat, not Crimson.

martypowell:
> Functionally, this looks great. @martypowell, per your comment above, we will look into a custom drop down icon as a backlog item -- not for this sprint.
> 
> One change since Sam finalized the design doc: The font for the dropdown text should be Montserrat, not Crimson.

At Dan, is this consistent for all form elements? I see that text input is also Crimson, i know you are probably annoyed at hearing this, but this should be consistent across form components.

danfox01:
> At Dan, is this consistent for all form elements? I see that text input is also Crimson, i know you are probably annoyed at hearing this, but this should be consistent across form components.

@martypowell, yes I discussed this with @sdurphy. He would like the selects to be Montserrat and the text inputs to be Crimson. Unless it poses a code challenge, let's handle that way for now and then reassess when we have a full form assembled and can see the the interplay between the elements.

martypowell:
> > At Dan, is this consistent for all form elements? I see that text input is also Crimson, i know you are probably annoyed at hearing this, but this should be consistent across form components.
> 
> @martypowell, yes I discussed this with @sdurphy. He would like the selects to be Montserrat and the text inputs to be Crimson. Unless it poses a code challenge, let's handle that way for now and then reassess when we have a full form assembled and can see the the interplay between the elements.

@danfox01 @sdurphy per sam, we are going go with Montserrat as 20px. @sgrg1 since i was messing with these values for sam, i just updated the code. 

Placeholder will remain, however, we will move towards using hints instead of placeholder text based on https://www.smashingmagazine.com/2018/06/placeholder-attribute/#a-solution. We all felt this was the best approach, but in some minor cases like the site search, we may still need the placeholder.

src/components/basic/form/Select.test.js
sgrg1: install prettier now to do the formatting

src/components/basic/form/Select.md
sgrg1: fixed

src/styles/partials/components/_input.scss
sgrg1: removed

danfox01:
This looks good to me. @jdomasky what do you think?

jdomasky:
@danfox1 I agree. This is good to go.

PR #96, opened 2019-08-28
Adjusted focus

martypowell:
To keep focus consistent we made the following changes

- Added 3px offset
- Changed from `vibrant blue` to `blue lightest`

@sdurphy @danfox01 please review, once it deploys.


danfox01:
@sdurphy I'm not 100% of this change. Can you verify it's good and if so, approve it?

martypowell:
@sdurphy I merged this, we can adjust tomorrow if needed.

Issue #36, opened 2019-08-08
Fieldset


martypowell:
**Please name the requested component**
Fieldset

**Please describe the requested component**
Use the default behavior based on https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset

**Please describe any required and optional properties for the component**

The fieldset should contain children of the type [FormField](https://github.com/baltimorecounty/dotgov-components/blob/integration/src/components/form/FormField.js)

**Example or Mockup**

Include an example of the component from another site, or a mockup
![image](https://user-images.githubusercontent.com/1933400/62708296-172a8500-b9c1-11e9-9c71-49c3d38e8a35.png)

**Acceptance Criteria**

**Rule Based**:

- Must follow UX standards
- Semantically correct - Must follow [W3C](https://www.w3.org/) standards
- Matches Mockup Styles

**Additional context**

n/a

martypowell:
This is dependent on all other form components being completed first.

martypowell:
this was completed in sprint 2

PR #100, opened 2019-09-03
Fieldset

theilman7:
#36 Creating pull request for fieldset. This adds styling and react markup for the FieldSet and Legend elements. This pull request also comes with a question. Should I address the <legend> styling of the <legend> elements previously submitted with the Checkbox Group and Radio Buttons appearing in the styleguide or should that be reserved for who submitted them?


sdurphy (commented):
This looks good to me, Travis.

src/components/basic/form/FieldSet.md
```diff
@@ -6,29 +6,55 @@
 import FieldSet from "./FieldSet";
 import Legend from "./Legend";
 import TextArea from "./TextArea";
```
martypowell: you only need to import the componetns you are using in the example.

martypowell (changes requested):
Can you also change the CheckedInputGroup to use the fieldset component example?

martypowell:
@theilman7 please also add a test, to ensure this component renders properly. See `CheckboxInputGroup.test.js` for an example

src/components/basic/form/FieldSet.js
```diff
@@ -1,5 +1,28 @@
 import React from "react";
+import PropTypes from "prop-types";
 
-const FieldSet = props => <fieldset>{props.children}</fieldset>;
+const FieldSet = props => {
```
martypowell, line 4: codacy wants this line to be

```js
const FieldSet = (props) =>
```

theilman7: Updated locally. Currently working through a terminal error when attempting to commit.

src/components/basic/form/FieldSet.md
theilman7: Updated locally. Currently working through a terminal error when attempting to commit.

src/components/basic/form/Legend.js
```diff
@@ -1,5 +1,5 @@
 import React from "react";
```
martypowell: instead of creating a legend component, i would put this directly in the fieldset component. and then make the prop type changes as we talked about on chat.

theilman7: Working on removing Legend component and integrating the <legend> tag directly in to the FieldSet. Will commit afterwards.

src/components/basic/form/FieldSet.js
theilman7: Updated and committed.

src/components/basic/form/Legend.js
theilman7: Updated and committed.

src/components/basic/form/FieldSet.md
theilman7: Updated and committed.

src/components/basic/form/FieldSet.js
martypowell: formatting rolled thsi back, which im ok with. I don't want to mess with this. The code review tools are just recommendations and I'll attempt to tweak.

src/components/basic/form/FieldSet.md
```diff
@@ -4,31 +4,53 @@
 
 ```jsx
 import FieldSet from "./FieldSet";
```
martypowell, line 6: not blocking, but you shouldn't have to import the actual component you are documenting, in this case fiedlset, react-styleguidist just knows.

src/components/basic/form/CheckedInputGroup.js
```diff
@@ -1,31 +1,29 @@
...
-			{hint && (
-				<span id={describedBy} className="dg_fieldset-hint">
-					{hint}
-				</span>
-			)}
-			{children}
-		</fieldset>
-	);
+  return (
+    <fieldset className="dg_fieldset" aria-describedby={describedBy}>
```
martypowell, line 8: @theilman7 I made a quick cahnge for you here, so we could get this completed. Since describedBy insn't requried, i set a default value for describedBy like this.

```js
const ariaDescribedBy = describedBy || title.replace(/\s/g, "");
```

martypowell: this just replaces all the spaces in the title

martypowell (dismissed):

martypowell:
@danfox01 @sdurphy this needs a stamp of approval.

danfox01:
@theilman7, looks great. One minor change to the example we provide — can you please remove the “1 - “ at the beginning of the legend text? It was on the comp but is a little confusing here.

theilman7:
@danfox01 @martypowell Removed the 1 - from the example string and pushed the commit up. The example copy now reads "Contact Information".

martypowell (approved):

martypowell:
@danfox01 ready for another look.

danfox01 (approved):

Issue #74, opened 2019-08-14
Alternate List


martypowell:
## Component Name

List

## Description

This component will expand upon the existing list component.

## Required and optional properties

This will follow the basic nature of an order or unordered list with some style adjustments.

## Example or Mockup

![image](https://user-images.githubusercontent.com/1933400/63040581-b9e87500-be93-11e9-9535-c6ffe11d46c8.png)

## Acceptance Criteria

**Rule Based**:

- Must follow UX standards
- Semantically correct - Must follow [W3C](https://www.w3.org/) standards
- Matches Mockup Styles

## Additional context

n/a

martypowell:
Completed in sprint 2.

PR #112, opened 2019-09-10
Alternate list

theilman7:
#74 Creating pull request for Alternate List component.

### DIFF
--- a/src/components/basic/lists/AlternateList.md
+++ b/src/components/basic/lists/AlternateList.md
@@ -1,0 +1,27 @@
+### Alternate List
+
+```jsx
+import ListItem from "./ListItem";
+<AlternateList>
+  <ListItem>Item One</ListItem>
+  <ListItem>Item Two</ListItem>
+  <ListItem>Item Three</ListItem>
+</AlternateList>;
+```
+
+Html Snippet:
+
+```html
+<ul class="dg_list">
+  <li>Item One</li>
+  <li>Item Two</li>
+  <li>Item Three</li>
+  <li>
+    Item Four
+    <ul>
+      <li>Sub item One</li>
+      <li>Sub item Two</li>
+    </ul>
+  </li>
+</ul>
+```
--- a/src/components/basic/lists/AlternateList.md
+++ b/src/components/basic/lists/AlternateList.md
@@ -3,10 +3,10 @@
 ```jsx
 import AlternateListItem from "./AlternateListItem";
 <AlternateList>
-  <AlternateListItem>Item One</AlternateListItem>
-  <AlternateListItem>Item Two</AlternateListItem>
-  <AlternateListItem>Item Three</AlternateListItem>
-  <AlternateListItem>Item Four</AlternateListItem>
+  <AlternateListItem link="#" title="Item One" />
+  <AlternateListItem link="#" title="Item Two" />
+  <AlternateListItem link="#" title="Item Three" />
+  <AlternateListItem link="#" title="Item Four" />
 </AlternateList>;
 ```
 
--- a/src/components/basic/lists/AlternateList.md
+++ b/src/components/basic/lists/AlternateList.md
@@ -15,16 +15,16 @@ Html Snippet:
 ```html
 <ul class="dg_alt-list">
   <li>
-    <a href="#">Item One<i class="fa fa-angle-right dg_i-float-right"></i></a>
+    <a href="#">Item One</a>
   </li>
   <li>
-    <a href="#">Item Two<i class="fa fa-angle-right dg_i-float-right"></i></a>
+    <a href="#">Item Two</a>
   </li>
   <li>
-    <a href="#">Item Three<i class="fa fa-angle-right dg_i-float-right"></i></a>
+    <a href="#">Item Three</a>
   </li>
   <li>
-    <a href="#">Item Four<i class="fa fa-angle-right dg_i-float-right"></i></a>
+    <a href="#">Item Four</a>
   </li>
 </ul>
 ```
--- a/src/components/basic/lists/AlternateList.md
+++ b/src/components/basic/lists/AlternateList.md
@@ -3,10 +3,10 @@
 ```jsx
 import AlternateListItem from "./AlternateListItem";
 <AlternateList>
-  <AlternateListItem link="#" title="Item One" />
-  <AlternateListItem link="#" title="Item Two" />
-  <AlternateListItem link="#" title="Item Three" />
-  <AlternateListItem link="#" title="Item Four" />
+  <AlternateListItem link="#" linkText="Item One" title="Link title" />
+  <AlternateListItem link="#" linkText="Item Two" title="Link title" />
+  <AlternateListItem link="#" linkText="Item Three" title="Link title" />
+  <AlternateListItem link="#" linkText="Item Four" title="Link title" />
 </AlternateList>;
 ```
 
@@ -15,16 +15,16 @@ Html Snippet:
 ```html
 <ul class="dg_alt-list">
   <li>
-    <a href="#">Item One</a>
+    <a href="#" title="Link title">Item One</a>
   </li>
   <li>
-    <a href="#">Item Two</a>
+    <a href="#" title="Link title">Item Two</a>
   </li>
   <li>
-    <a href="#">Item Three</a>
+    <a href="#" title="Link title">Item Three</a>
   </li>
   <li>
-    <a href="#">Item Four</a>
+    <a href="#" title="Link title">Item Four</a>
   </li>
 </ul>
 ```
--- a/src/components/basic/lists/AlternateList.md
+++ b/src/components/basic/lists/AlternateList.md
@@ -1,27 +1,30 @@
-### Alternate List
+### Unordered List
 
 ```jsx
-import ListItem from "./ListItem";
+import AlternateListItem from "./AlternateListItem";
 <AlternateList>
-  <ListItem>Item One</ListItem>
-  <ListItem>Item Two</ListItem>
-  <ListItem>Item Three</ListItem>
+  <AlternateListItem>Item One</AlternateListItem>
+  <AlternateListItem>Item Two</AlternateListItem>
+  <AlternateListItem>Item Three</AlternateListItem>
+  <AlternateListItem>Item Four</AlternateListItem>
 </AlternateList>;
 ```
 
 Html Snippet:
 
 ```html
-<ul class="dg_list">
-  <li>Item One</li>
-  <li>Item Two</li>
-  <li>Item Three</li>
+<ul class="dg_alt-list">
   <li>
-    Item Four
-    <ul>
-      <li>Sub item One</li>
-      <li>Sub item Two</li>
-    </ul>
+    <a href="#">Item One<i class="fa fa-angle-right dg_i-float-right"></i></a>
+  </li>
+  <li>
+    <a href="#">Item Two<i class="fa fa-angle-right dg_i-float-right"></i></a>
+  </li>
+  <li>
+    <a href="#">Item Three<i class="fa fa-angle-right dg_i-float-right"></i></a>
+  </li>
+  <li>
+    <a href="#">Item Four<i class="fa fa-angle-right dg_i-float-right"></i></a>
   </li>
 </ul>
 ```
--- a/src/components/basic/lists/AlternateListItem.js
+++ b/src/components/basic/lists/AlternateListItem.js
@@ -1,12 +1,26 @@
 import React from "react";
+import PropTypes from "prop-types";
 
-const AlternateListItem = props => (
-  <li>
-    <a href="#">
-      {props.children}
-      <i className="fa fa-angle-right dg_i-float-right" aria-hidden="true"></i>
-    </a>
-  </li>
-);
+const AlternateListItem = props => {
+  const { link, title } = props;
+  return (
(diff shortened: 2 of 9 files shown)


src/components/basic/lists/AlternateListItem.js
```diff
@@ -0,0 +1,12 @@
+import React from "react";
+
+const AlternateListItem = props => (
```
martypowell: This should have a prop for `href` that is required (using proptypes). You should probably also have other link properties on this.

src/components/basic/lists/AlternateListItem.js
```diff
@@ -0,0 +1,12 @@
...
+  <li>
+    <a href="#">
+      {props.children}
+      <i className="fa fa-angle-right dg_i-float-right" aria-hidden="true"></i>
```
martypowell: I think in this case, we probably want to achieve the arrow with css, since it is just decoration.

src/styles/partials/components/_alt-list.scss
```diff
@@ -0,0 +1,56 @@
...
+
+ul.dg_alt-list {
+  list-style-type: none;
+  margin-bottom: $margin-large;
+  margin-left: 0;
+  padding-left: 0;
+  text-decoration: none;
+}
+
+.dg_i-float-right {
```
martypowell: once changed to use will be positioned rather than floated.

src/styles/partials/components/_alt-list.scss
```diff
@@ -0,0 +1,56 @@
...
+  color: $gray-light;
+  float: right;
+  font-size: $large-icon-size;
+}
+
+.dg_alt-list {
+  li {
+    border-bottom: $border-width-small solid $gray-light;
+    margin-top: $default-margin;
+    -webkit-transition: all 0.3s ease 0s;
```
martypowell: i don't think we want to transition everything, we will only want to transition what already is handled by the a and button transitions. You should also check out the _animations.scss mixins file. You might have questions, if so, please let me know.

src/styles/partials/components/_alt-list.scss
```diff
@@ -0,0 +1,56 @@
...
+    a {
+      color: $blue;
+      font-family: $default-heading-font;
+      font-size: $smallest-heading-font-size;
+      font-weight: $font-bold;
+      line-height: 30px;
+      text-decoration: none;
+      text-transform: uppercase;
+    }
+    &:hover {
```
martypowell: you prlbably want to include this block above the a. Reason being this directly applies to the li, where a is a child.

src/styles/partials/components/_alt-list.scss
```diff
@@ -0,0 +1,56 @@
...
+      border-bottom: $border-width-small solid $vibrant-blue;
+      cursor: pointer;
+      margin-top: $default-margin;
+      text-decoration: none;
+      > a {
```
martypowell: I think you can combine `a` and `> a`?

martypowell (changes requested):
Please take a look at the comments.

martypowell:
@theilman7 be sure to do a `git pull` before making these changes.

src/components/basic/lists/AlternateListItem.js
theilman7: Updated locally. Will commit after the tests run and pass.

src/styles/partials/components/_alt-list.scss
theilman7: Moved all of the block up to apply to the li, with the exception of the text-decoration property. The <a> tag needs to be targeted with the text-decoration tag directly in order to remove default link styles. Adding text-decoration to the parent, li, will not remove default link styling, such as underline. Will upload with a new commit shortly.

src/styles/partials/components/_alt-list.scss
theilman7: Taking a look at what exists in the _animations.scss file and will follow up with any questions.

theilman7: Updating with the mixin and will make a commit now.

src/components/basic/lists/AlternateListItem.js
theilman7: Committed initial proptypes.

src/components/basic/lists/AlternateListItem.js
theilman7: Working on this now. Will commit when finished.

src/styles/partials/components/_alt-list.scss
theilman7: Working on this as well, since it goes hand-in-hand with the previous update.

jdomasky:
@theilman7 @martypowell First round of testing on this component is complete. Two issues were found and confirmed with @sdurphy and @danfox01:

1. **Height and vertical alignment of list items:**
In the mockup, each list item has a height of 46px. The text is 17px from the bottom border. The text and arrow are vertically aligned in the middle of the list item container.
In the Now rendering, each list item has a height of 55px. The text is 9px from the bottom border. The text and arrow are vertically aligned close to the bottom of the list item container.
Action to be taken: The rendering should be changed to match the mockup.

2. **Bottom border of list item (hovered):**
In the mockup, when the bottom border of a list item is in a hover condition, the color doesn't change.
In the Now rendering, the bottom border color changes upon hover to vibrant-blue.
Action to be taken: The bottom border shouldn't change. The color change should be disabled.

src/components/basic/lists/AlternateListItem.js
theilman7: Submitted via the latest commit.

theilman7:
@jdomasky Just finished addressing item two (bottom border of list item hovered). I will begin working to correct the other issue and will follow up when they are each ready to review.

theilman7:
@jdomasky  Updates addressing feedback from both items should be ready for review in the most recent commit. Let me know when you have a chance to review.

src/components/basic/lists/AlternateList.md
```diff
@@ -0,0 +1,30 @@
...
+  <AlternateListItem link="#" title="Item Four" />
+</AlternateList>;
+```
+
+Html Snippet:
+
+```html
+<ul class="dg_alt-list">
+  <li>
+    <a href="#">Item One<i class="fa fa-angle-right dg_i-float-right"></i></a>
```
martypowell: @theilman7  it looks like you are specifying a title, but its not in the snippet. either maybe this cahnged, and is just out of date, or it's not rendering?

src/styles/partials/components/_alt-list.scss
```diff
@@ -0,0 +1,54 @@
+ol.dg_alt-list {
+  list-style-type: none;
+  margin-bottom: $margin-large;
+  margin-left: 0;
+  padding-left: 0;
+  text-decoration: none;
+}
+
+ul.dg_alt-list {
```
martypowell: i actually think you can jsut make this.

```scss
.dg_alt-list {
  list-style-type: none;
  margin-bottom: $margin-large;
  margin-left: 0;
  padding-left: 0;
  text-decoration: none;
}
```

Shouldn't need to do the ol.ul. stuff

src/styles/partials/components/_alt-list.scss
```diff
@@ -0,0 +1,54 @@
...
+  list-style-type: none;
+  margin-bottom: $margin-large;
+  margin-left: 0;
+  padding-left: 0;
+  text-decoration: none;
+}
+
+.dg_alt-list {
+  li {
```
martypowell, line 7: With the comment above you should be able to take all the nested rules and move them up.

src/styles/partials/components/_alt-list.scss
```diff
@@ -0,0 +1,54 @@
...
+    margin-top: $default-margin;
+    text-decoration: none;
+    text-transform: uppercase;
+    @include default-transition(color);
+    a {
+      text-decoration: none;
+
+      &:after {
+        color: $gray-light;
+        content: "\f105";
```
martypowell: please add a comment describing the icon here. 

```scss
/** fa-arrow or wahtever, and if its not obvious describe it in english**/
```

src/styles/partials/components/_alt-list.scss
```diff
@@ -0,0 +1,54 @@
...
+        float: right;
+        font-family: "Font Awesome 5 Pro";
```
martypowell: there should be a variable for this one. It might also be worth making an icon mixin if you are up for it, i know we are repeating

```scss
content: 'icon caharters';
font-family: 'font awesome';
```

in a couple different places now.

src/styles/partials/components/_alt-list.scss
theilman7: Removed ul and ol items and condensed under .dg_alt-list.

src/styles/partials/components/_alt-list.scss
theilman7: Items have been nested as per previous comment.

src/styles/partials/components/_alt-list.scss
theilman7: Added a descriptive comment and committed.

src/styles/partials/components/_alt-list.scss
theilman7: Created and committed a mixin for this.

src/components/basic/lists/AlternateList.md
theilman7: Updated the HTML snippet to remove the markup for the icon, as it is being included at the css level now. Regarding the title, this is represented as the text reading Item One, Item Two etc. in the html snippet. I might be misunderstanding the comment though.

src/components/basic/lists/AlternateListItem.js
```diff
@@ -0,0 +1,20 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+const AlternateListItem = props => {
+  const { link, title } = props;
+  return (
+    <li>
+      <a href={link}>{title}</a>
```
martypowell: @theilman7 so there is a title for an `a` and there there is the `linkText`. I think i probably commented this, but didn't make it clear.

martypowell: linkText would be required, in addition to the "title"

src/styles/partials/components/_alt-list.scss
theilman7: Condensed the hover styles for the > a selector to a. This has been updated and submitted with the latest commit.

src/styles/mixins/_icon-content.scss
```diff
@@ -0,0 +1,5 @@
+@mixin right-arrow-icon {
```
martypowell: i expected this to be more like

```scss
/* $font-awesome-unicode can be found in there docs, the format is '\f###' example '\f105' */
@mixin icon-content($font-awesome-unicode) {
  content: $font-awesome-unicode;
  font-family: $font-awesome-pro;
}
```

**Usage**
```scss

@include content-icon("\f105");
```

This way we can re-use it for any icon.

theilman7: Updated and committed the mixin.

src/components/basic/lists/AlternateListItem.js
theilman7: Updated with the most recent commit.

src/styles/mixins/_icon-content.scss
```diff
@@ -0,0 +1,5 @@
+@mixin icon-content($font-awesome-unicode) {
+  /** Adds an icon from FontAwesome. Formatting and unicode references can be found at fontawesome.com. Formatting will roughly appear as \f### **/
```
martypowell: I would just move this comment above the mixin, and this is the typical location in most programming languages for functions.

src/components/basic/lists/AlternateListItem.js
```diff
@@ -0,0 +1,24 @@
+import React from "react";
```
martypowell, line 1: @theilman7 I lied, can you create a test file for this please. Minimal render test is fine.

src/styles/mixins/_icon-content.scss
theilman7: Sounds good. Just moved the comment above the mixin and committed.

jdomasky:
> 
> 
> @jdomasky Updates addressing feedback from both items should be ready for review in the most recent commit. Let me know when you have a chance to review.

@theilman7 I see that this component is still changing. Let me know when I should proceed with the retest.

martypowell (approved):

martypowell:
@danfox01 @sdurphy this is just pending the rubber stamp from you guys.

src/components/basic/lists/AlternateListItem.test.js
```diff
@@ -0,0 +1,26 @@
...
+    </AlternateList>,
+    div
+  );
+  ReactDOM.unmountComponentAtNode(div);
+});
+
+it("renders an `ol` without crashing", () => {
+  const div = document.createElement("div");
+  ReactDOM.render(
+    <AlternateList type="ordered">
```
martypowell, line 20: actually @theilman7 , you don't need the list container, but not blocking at this point, we can cleaup later.

martypowell: woudl just be

"render without crashing"

<AlternateListItem ... />

sdurphy (approved):

Issue #62, opened 2019-08-12
Decorative Button Group


martypowell:
## Component Name

DecorativeButtonGroup

## Description

This component is a group of decorative buttons. Decorative buttons (created in a previous sprint) consist of a round icon and a label. 

This component is used on the home page as a stylistic alternative to button-cards.
## Required and optional properties

Each list item will have:

- **icon** - type - _string_ - font awesome icon
- **text** - type - _string_ - The link text to be displayed below the icon
- **link** - type - string - The url the button should navigate the user to

## Example or Mockup

![image](https://user-images.githubusercontent.com/1933400/62875401-f760cd00-bcf0-11e9-9868-3d5bf43d837a.png)

A smaller list with not text

![image](https://user-images.githubusercontent.com/1933400/63040691-fc11b680-be93-11e9-8047-dfb89805fa26.png)

## Acceptance Criteria

Please include in the following format. In some cases you will not need both types.

**Rule Based**:
- Buttons in a button group should maintain all existing functionality (TODO: Does this make sense?)
- Buttons should respond to breakpoints in the following ways:
  - Display 4-up at 992px and up.
  - Display 2-up at 768-991px.
  - Display in a single column at screens smaller than 768px.
- Must follow UX standards
- Semantically correct - Must follow [W3C](https://www.w3.org/) standards
- Matches Mockup Styles

## Additional context

We might be able to use to use [ButtonList](https://github.com/baltimorecounty/dotgov-components/issues/61) and style the alternate version with css.

martypowell:
We ended up using the grid to handle this, so the group was unneeded in this occasion.

Issue #61, opened 2019-08-12
Card-Button Group


martypowell:
## Component Name

CardButtonGroup

## Description

This component groups [IconButtons](https://github.com/baltimorecounty/dotgov-components/issues/13). The component can be configured for various uses throughout the site, such as accommodating different numbers of card-buttons per row, or appearing on light vs. dark backgrounds.

## Required and optional properties

Children will be icon buttons.

## Example or Mockup

_Default_

![image](https://user-images.githubusercontent.com/1933400/62874909-072be180-bcf0-11e9-9d35-4eb4172e3a6a.png)

_Alternate_

![image](https://user-images.githubusercontent.com/1933400/62874772-cf249e80-bcef-11e9-87fe-4d935a2aa8b3.png)

## Acceptance Criteria

Please include in the following format. In some cases you will not need both types.

### Rule Based

- Web services team members must be able to easily configure the number of cards that are displayed per row when viewed at full screen, so they can choose the best layout for the space available. 
  - For example, the homepage shows rows of 5 items by default. Subpage examples show rows of 3, in a smaller space. 3 vs. 5 items should be configurable.
- Regardless of the desktop configuration, cards must size down and break gracefully at all breakpoints, depending on the bounds of their parent container. The max number of cards per container size are:
  - more than 1000px: 5 cards
  - 576-999px: 3 cards
  - less than 576px: 1 card
- Must follow UX standards
- Semantically correct - Must follow [W3C](https://www.w3.org/) standards
- Matches Mockup Styles

## Additional context

tbd

martypowell:
@danfox01 we will responsive break points for these lists.

martypowell:
this was completed as part of sprint 2

PR #113, opened 2019-09-11
Card Button Group

martypowell:
Addresses #61 

NOTE: This does not display properly in the style guide because the layout in our site is different than the styleguide which messes with the breakpoints. This will need to be tested in the prototype. It is possible that this may need to be tweaked as part of #62

### DIFF
--- a/src/components/basic/IconLink.md
+++ b/src/components/basic/IconLink.md
@@ -98,5 +98,93 @@ Html Snippet:
 </a>
 ```
 
+### Grid
+
+```jsx
+<div className="dg_main-content">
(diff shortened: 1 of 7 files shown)


martypowell:
as confirmed with @danfox01 and @sdurphy we will handle heading margins as part of this pr, working on that now.

danfox01:
@martypowell do I have anything to test here?

martypowell:
> @martypowell do I have anything to test here?

@danfox01 No, everything was added, well have to verify in the prototype, since the styleguide doesn't work at the same break points.

PR #116, opened 2019-09-13
Exported beta components

martypowell:
I keep forgetting that we need to export theses components for the npm packages to work. This exports all of the components from the current milestone.


tmccoy529 (approved):
Approved with excrement.....or excitement....stupid autocorrect

PR #124, opened 2019-09-16
added padding to images

tmccoy529:
Added padding to images with small break.

### DIFF
--- a/src/styles/partials/components/_emphasized-image.scss
+++ b/src/styles/partials/components/_emphasized-image.scss
@@ -55,8 +55,11 @@
 }
 
 @media screen and (max-width: $small-breakpoint) {
-  p:first-of-type,
+  p:first-of-type {
+    margin: $base-spacing-unit * 3 auto;
+  }
   p:last-of-type {
+    margin-bottom: $base-spacing-unit * 9, auto;
     text-align: center;
   }
 }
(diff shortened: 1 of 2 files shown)


src/styles/partials/components/_emphasized-image.scss
```diff
@@ -55,8 +55,11 @@
 }
 
 @media screen and (max-width: $small-breakpoint) {
-  p:first-of-type,
+  p:first-of-type {
```
martypowell: A couple things

- i would leave it as and override the margin for last of type.
- you are missing .dg_card_content this would apply to p in this case (i think this is integration, so you might just be able to pull the latest and merge)
- confirmed with @sdurphy margin-large should be used here

Think something like this is how i would solve

```scss
.dg_card-content {
    p:first-of-type,
    p:last-of-type {
      margin: $base-spacing-unit * 3 auto;
      text-align: center;
    }

    p:last-of-type {
      margin-bottom: $margin-large;
    }
  }
```

martypowell (changes requested):
Sorry, this probably felt super easy.

src/styles/partials/components/_emphasized-image.scss
```diff
@@ -55,8 +55,11 @@
...
+    margin: $base-spacing-unit * 3 auto;
+  }
   p:last-of-type {
+    margin-bottom: $base-spacing-unit * 9, auto;
```
martypowell: comma here should throw an error when you build the style sheet.

martypowell (approved):

tmccoy529:
@martypowell hahaha it would have been if I had been paying attention and didn't have an annoying bird chirping in my ear while I'm trying to fix this.

PR #129, opened 2019-09-16
Footer

theilman7:
Adding corresponding SCSS for the [Footer pull request](https://github.com/baltimorecounty/dotgov-prototype-2.0/pull/16).


src/styles/partials/components/_footer.scss
```diff
@@ -0,0 +1,94 @@
...
+      float: left;
+      margin-right: 5%;
+      width: 40%;
+    }
+  }
+  .dg_seal {
+    max-width: 100px;
+  }
+}
+.dg_social-icon_wrapper {
```
martypowell: can we call this __container to be consistent please?

src/styles/partials/components/_footer.scss
```diff
@@ -0,0 +1,94 @@
...
+  margin-bottom: $default-margin;
+  margin-top: $margin-largest;
+}
+
+.dg_lg_gray {
+  color: $gray;
+  font-size: $large-icon-size;
+}
+
+@media (max-width: 767px) {
```
martypowell: you can use $medium-breakpoint here.

src/styles/partials/components/_footer.scss
```diff
@@ -0,0 +1,94 @@
...
+  .dg_brand-footer {
+    li {
+      width: 100% !important;
```
martypowell: we do not want to use !important unless we really need to. Does this work without it?

src/styles/partials/components/_footer.scss
```diff
@@ -0,0 +1,94 @@
...
+.dg_social-icon_wrapper {
+  margin-bottom: $margin-largest;
+  .dg_icon-link.dg_icon-link--small {
+    i {
+      font-size: $default-icon-size;
+    }
+  }
+}
+
+.dg_navigation-footer {
```
martypowell: .dg_navigation-footer and .dg_social-footer share the styles, so they can be combined here.

theilman7: Will update and re-submit to address the feedback.

src/styles/partials/components/_footer.scss
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  p {
+    padding-bottom: $padding-largest;
+  }
+
+  a {
+    text-decoration: none;
+  }
+  .dg_alt-list {
```
martypowell: I'm a little nervous of what this is doing, gonan pull up the prototype and check this out. ideally we wouldnt' be doign anything to this list.

src/styles/partials/components/_footer.scss
theilman7: Updated and committed.

src/styles/partials/components/_footer.scss
theilman7: Updated and committed.

src/styles/partials/components/_footer.scss
theilman7: Since this overrides a previous width style with the media query, it will need !important to work.

src/styles/partials/components/_footer.scss
theilman7: This has been condensed and submitted with a commit. I condensed these two classes to one class, .dg_footer-list-column. Let me know if this works or if you would like it tackled differently.

martypowell (dismissed):

PR #133, opened 2019-09-18
Refactor Logotype to be Reusable in Footer

martypowell:
Purpose of this pr is to make the "Brand Text" resuable in the header and footer. The header text is link, while the footer text is not.

- Removed link from `LogoType` and moved it to the `SiteHeader`
- Added title and fixed href in `SiteHeader`
- Renamed `LogoType` to `BrandText` to make sure more sense to both header and footer.
- Exported missing component `Fieldset`


src/styles/partials/components/_brand-text.scss
```diff
@@ -1,4 +1,4 @@
-.dg_logotype {
+.dg_brand-text {
   border-right: $border-width-small solid $gray-light;
   padding: 0 $padding-large 0 $padding-largest;
   text-decoration: none;
```
martypowell: this can be removed (i think).

martypowell: removed.

tmccoy529 (dismissed):
This looks good to me

martypowell:
No version, bump, will do now.

PR #144, opened 2019-09-27
Images

sgrg1:
Images component


sgrg1:
@martypowell , this image does not have the latest code of summaylist that is why deepscan is showing error message . Let me know what you want me to do.

martypowell:
> @martypowell , this image does not have the latest code of summaylist that is why deepscan is showing error message . Let me know what you want me to do.

It seems like you branched off of your SummaryList branch instead of the integration branch. There are lots of ways we could approach this, but I'll review this as is for now. In the meantime, there are requested changes on that pr. Once that one is completed we can merge the two branches.

src/components/site/Images.md
```diff
@@ -0,0 +1,22 @@
...
+<Images
+  imageSrc="//baltimorecountymd.gov/sebin/t/t/homepage-county-executive.jpg"
+  content="Lorem Ipsum is simply dummy text of the printing and typesetting of Lorem Ipsum.industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book. It has survived not only five"
+></Images>
+```
+
+Html Snippet:
+```html
+<div>
+<div class="dg_image">
```
martypowell: the dg_image class shouldn't be on a div as the Site Executive users won't be creating a div that wraps around an image when they are using the cms.

src/components/site/Images.md
```diff
@@ -0,0 +1,22 @@
...
+  <img src="//baltimorecountymd.gov/sebin/t/t/homepage-county-executive.jpg" aria-hidden="true"></div>
```
martypowell: In this case, we actually don't want to use [aria-hidden](https://www.w3.org/WAI/PF/aria/states_and_properties#aria-hidden)="true" because this image is actually visible. We will also want to include an alt tag. This [link](https://www.commonplaces.com/blog/writing-alt-tags-for-images/) has a nice explanation of what the alt tag does.

src/components/site/Images.md
martypowell: in this case, we want default functionality here, so we probably don't even need to add a class to this image, and rather do something like this in the css

```scss
#main-content img
```

this targets images inside of the content section of the page, and won't impact the images on the rest of the site.

src/components/site/Images.md
```diff
@@ -0,0 +1,22 @@
...
+  <div class="dg_image left">
```
martypowell: Please move this class to the image, and remove the div.

martypowell: add `alt` tag and remove `aria-hidden`

src/components/site/Images.md
```diff
@@ -0,0 +1,22 @@
...
+    <img src="//baltimorecountymd.gov/sebin/t/t/homepage-county-executive.jpg" aria-hidden="true">
+  </div>
+  <p>Lorem Ipsum is simply dummy text of the printing and typesetting of Lorem Ipsum.industry. Lorem Ipsum has been the industry's standard dummy text ever   since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book. It has survived not only five</p>
+  <div class="dg_image right">
```
martypowell: Please move this class to the image, and remove the div. add alt tag and remove aria-hidden

src/index.js
```diff
@@ -7,6 +8,7 @@ export { default as Section } from "./components/basic/containers/Section";
 /** Form Components */
 export { default as Checkbox } from "./components/basic/form/Checkbox";
 export { default as CheckedInput } from "./components/basic/form/CheckedInput";
+export {default as  Images} from "./components/site/Images";
```
martypowell: We won't need to export this component, as we we handle this in the prototype as a page.

src/styles/partials/components/_images.scss
```diff
@@ -0,0 +1,42 @@
...
+  display: block;
+  position: relative;
+  img {
+    display: block;
+    object-fit: fill;
+    width: 100%;
+  }
+}
+
+.left {
```
martypowell: `.left` and `.right` should next below `.dg_image`

src/styles/partials/components/_images.scss
```diff
@@ -0,0 +1,42 @@
+.dg_image {
```
martypowell: this should be nested below `#main-content`

src/components/site/Images.js
```diff
@@ -0,0 +1,32 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+const Images = props => {
```
martypowell: If we want to keep this in the design system, You can just lines 8 - 20 directly to a Images.md file and then update the [styleguide.config.js](https://github.com/baltimorecounty/dotgov-components/blob/integration/styleguide.config.js#L46) by adding a section, named "Images" under Site Elements. See [here](https://github.com/baltimorecounty/dotgov-components/blob/integration/styleguide.config.js#L9) for an example.

src/components/site/Images.md
sgrg1: done

src/index.js
sgrg1: done

src/styles/partials/components/_images.scss
sgrg1: done

src/styles/partials/components/_images.scss
sgrg1: done

src/components/site/Images.js
sgrg1: @martypowell , I did understand this part . please help

martypowell: resolved per our conversation available here https://github.com/baltimorecounty/dotgov-components/commit/9077c47f7eca6f4b4296d2ad55a87eb142d71f5d

src/styles/partials/components/_images.scss
martypowell: sorry, nested in `.dg_image`

```scss
.dg_image {
  display: block;
  ...

  .left {
     something-goes: here;
  }
}
```

martypowell: same with .right

src/styles/partials/components/_images.scss
```diff
@@ -0,0 +1,45 @@
...
+    width: 100%;
+  }
+  .dg_image {
+    display: block;
+    position: relative;
+  }
+}
+.left {
+  float: left;
+  margin-top: $margin-small;
```
martypowell: margin-top, margin-bottom and max-width are the same for left and right, so we should combine those into one block

```scss
.left,
.right {
  /* shared rules between .left and .right*/
}

.left {
/* only rules specific to .left */
}

.right {
/* only rules specific to .right */
}

File: src/styles/partials/components/_images.scss [hunk 10]
```diff
@@ -0,0 +1,45 @@
...
+    position: relative;
+    display: block;
+  }
+}
+
+@media screen and (max-width: $medium-breakpoint) {
+  .dg_image {
+    max-width: 100%;
+    position: relative;
+    display: block;
```
martypowell: lines 40 and 41 already declare the position and display, so it is redundant, as you move down the css file. These can be removed.

martypowell (changes requested):
See the other comments, it looks like the top image could use margin-bottom: $padding-largest; to match the spacing of the other images.

src/styles/partials/components/_images.scss
sgrg1: done

src/styles/partials/components/_images.scss
sgrg1: done

src/components/site/Images.md
```diff
@@ -0,0 +1,63 @@
...
+    since the 1500s, when an unknown printer took a galley of type and scrambled
+    it to make a type specimen book. It has survived not only five
+  </p>
+</div>
+```
+
+Html Snippet:
+
+```html
+<div class="dg_image-content">
```
martypowell: this div needs to change to `<div id="dg_main-content"><!-- stuff goes here --> </div>`. This denote the main content section of the application.

The css will have to change accordingly

src/styles/partials/components/_images.scss
```diff
@@ -0,0 +1,43 @@
+.dg_image-content {
```
martypowell: this will change to `#dg_main-content`

src/styles/partials/components/_images.scss
```diff
@@ -0,0 +1,43 @@
...
+  img {
+    display: block;
+    object-fit: fill;
```
martypowell: this can be removed, as it's not doing anything.

martypowell (changes requested):
A couple quick fixes

src/components/site/Images.md
sgrg1: done

src/styles/partials/components/_images.scss
sgrg1: done

src/styles/partials/components/_images.scss
sgrg1: done

martypowell (approved):
@sgrg1 looks good, merging on behalf you momentarily.

PR #151, opened 2019-10-01
Popular service padding#14954

sgrg1:
Make the Popular Services top and bottom padding 30px at all breakpoints.Bug 14954


martypowell:
@sgrg1 this collided with #159 so the spacing has already been addressed. It just happened that I glanced at that one first. Therefore closing this.

PR #154, opened 2019-10-03
Accordion comp

tmccoy529:
Added the accordion example to the style guide. This has light back ground setup so all selectable panels can stay open together and a dark background where the accordion is setup to not allow multiple selectable panels opened at once.


sdurphy:
@tmccoy529 what URL should I be using to review this?

tmccoy529:
@sdurphy this PR was supposed to generate a deployment URL for that but looks like it failed to do it. Sooo this looks like a dead-end.

src/styles/partials/components/_accordion-dark.scss
```diff
@@ -0,0 +1,28 @@
+.dg_accordion.dg_section.dark {
+  .dg_accordion-btn:focus {
+    color: $white;
+  }
+  .dg_allitems {
+    border-bottom: 1px $white solid;
```
theilman7: Recommend changing 1px to use the sass variable: $border-width-small.

src/styles/partials/components/_accordion-dark.scss
```diff
@@ -0,0 +1,28 @@
...
+    color: $white;
+    &:hover {
+      color: $blue-lightest;
+    }
+  }
+  .dg_accordion__collapsible {
+    color: $white;
+    border-bottom: 1px $white solid;
```
theilman7: Recommend changing 1px to use the sass variable: $border-width-small.

src/styles/partials/components/_accordion.scss
```diff
@@ -0,0 +1,74 @@
+.dg_accordion__collapsible {
+  color: $blue;
+  font-family: $default-heading-font;
+  font-size: $h2-font-size;
+  font-weight: $font-bold;
+  width: 100%;
+  border-bottom: 1px $black solid;
```
theilman7: Recommend changing 1px to use the sass variable: $border-width-small.

src/styles/partials/components/_accordion.scss
```diff
@@ -0,0 +1,74 @@
...
+      right: 0;
+    }
+  }
+}
+
+.dg_button-link.dg_allitems {
+  padding-bottom: $padding-small;
+  padding-right: $base-spacing-unit;
+  text-align: right;
+  border-bottom: 1px $black solid;
```
theilman7: Recommend changing 1px to use the sass variable: $border-width-small.

src/styles/partials/components/_accordion-dark.scss
tmccoy529: Fixed

src/styles/partials/components/_accordion-dark.scss
tmccoy529: Fixed

src/styles/partials/components/_accordion.scss
tmccoy529: Fixed

src/styles/partials/components/_accordion.scss
tmccoy529: Fixed

martypowell:
@tmccoy529 looks like the container padding is not right.

image

I can't tell exactly what it is from the xd doc, but lets try default padding for now.

martypowell:
@tmccoy529 think the button should have it's cursor set to pointer, see https://design-system.service.gov.uk/components/accordion/

tmccoy529:
@martypowell updated both of those changes. Default padding and button now has cursor pointer.

martypowell:
@tmccoy529 looks like the focus is slightly changed from the default, so what you'll have to do is override the default.

image

So in this case you would want to set .dg_accordion-btn:focus outline to none. Add a span around the button text with a meaningful class name. Then add the default focus styles to that span.

I know this is probably confusing so if you need help let me know. Your final results should look like this:

image

src/styles/partials/components/_accordion.scss
```diff
@@ -0,0 +1,74 @@
...
+    &:focus {
+      color: $blue;
+      outline: none;
+    }
+    &:after {
+      position: absolute;
+      padding-top: $padding-small;
+      padding-bottom: $padding-small;
+      font-size: $h1-font-size;
+      @include icon-content("\f107"); /*Font awesome right arrow*/
```
martypowell: can you please move "\f107" to a variable in the "_icon.scss" file as $arrow-right? Same would go for the down arrow. I know we are using these arrows in the breadcrumbs and footer, so it would be good to control them in one place if we decide to change them.

martypowell (changes requested):

src/components/site/Accordion.js
```diff
@@ -0,0 +1,71 @@
...
+  );
+};
+
+Accordion.propTypes = {
+  /** Populates the selectable items in the accordion */
+  collapseItems: PropTypes.array.isRequired,
+  /** Content of the selected item */
+  toggleMultiOpen: PropTypes.string,
+  /** Can add extra cssclass to the element EX. Dark */
+  className: PropTypes.string
```
martypowell: this is unneeded as it's available on all react elements.

src/components/site/Accordion.js
```diff
@@ -0,0 +1,71 @@
...
+      ))}
+    </div>
+  );
+};
+
+Accordion.propTypes = {
+  /** Populates the selectable items in the accordion */
+  collapseItems: PropTypes.array.isRequired,
+  /** Content of the selected item */
+  toggleMultiOpen: PropTypes.string,
```
martypowell: per our chat, please make this a boolean then add the appropriate class. `canHaveMultiplePanelsOpen`, maybe?

src/components/site/Accordion.md
```diff
@@ -0,0 +1,170 @@
...
+  },
+  {
+    id: 14,
+    header: "Accordion Item #4",
+    subheader: "",
+    text: <a href="https://www.w3schools.com">This is a link </a>
+  }
+];
+
+<Accordion className="dg_section dark" collapseItems={collapseItems} />;
```
martypowell: This needs to change slightly because of how web services will use this component. Web service will dump this in a section. That section may be white, dark or light. The accordion styles should change based on that section. So in this case your example would like.

```jsx
<Section className="dark">
  <Accordion collapseItems={collapseItems} />
</Section>
```

This will require you to change your css a bit.

martypowell: if you have questions about this, let me know.

src/styles/partials/components/_accordion.scss
tmccoy529: this is done

src/components/site/Accordion.js
tmccoy529: done

src/components/site/Accordion.js
tmccoy529: done

src/components/site/Accordion.md
tmccoy529: this should be done

src/styles/partials/components/_accordion.scss
```diff
@@ -0,0 +1,74 @@
...
+      top: 0;
```
martypowell: based on the icon change, i believe this needs to be 8px. We will have to confirm with design.

src/styles/partials/components/_accordion.scss
```diff
@@ -0,0 +1,74 @@
...
+      right: 0;
+    }
+  }
+}
+
+.dg_button-link.dg_allitems {
+  padding-bottom: $padding-small;
+  padding-right: $base-spacing-unit;
+  text-align: right;
+  border-bottom: $border-width-small $black solid;
```
martypowell: per our chat, this can be moved to the list first-child, you shouldn't need the width: 100% there as its already displaying block.

src/components/site/Accordion.js
```diff
@@ -0,0 +1,76 @@
...
+      ))}
+    </div>
+  );
+};
+
+Accordion.propTypes = {
+  /** Populates the selectable items in the accordion */
+  collapseItems: PropTypes.array.isRequired,
+  /** Content of the selected item */
+  toggleMultiOpen: PropTypes.bool
```
martypowell: can we change the name of this to something that is more bool like? `canHaveMultiplePanelsOpen` this is all i can thinnk of so there is probably a better name, other prefixes could be `has` or `is`.  as seen here, https://hackernoon.com/the-art-of-naming-variables-52f44de00aad

src/components/site/Accordion.js
```diff
@@ -0,0 +1,76 @@
...
+          </div>
+        </div>
+      ))}
+    </div>
+  );
+};
+
+Accordion.propTypes = {
+  /** Populates the selectable items in the accordion */
+  collapseItems: PropTypes.array.isRequired,
```
martypowell: I think maybe just describing what properties a collapsed item might have, could be good here.

src/components/site/Accordion.js
```diff
@@ -0,0 +1,76 @@
...
+      {/* <div className="dg_accordion_buttonall-container "> */}
+      <button
+        className="dg_allitems dg_button-link"
+        key="collpaseAllActionButton1"
+      >
+        Open All
+      </button>
+      {/* </div> */}
+
+      {collapseItems.map(item => (
```
martypowell: this would be a good case for destructing the `item`.  This would give a good idea, of waht propeties exist on an item.

example

```jsx
const { id, subheader, text } = item;
```

and any other properties i missed.

src/components/site/Accordion.md
```diff
@@ -0,0 +1,178 @@
...
+    id: 2,
+    header: "Accordion Item #2",
+    subheader: "",
+    text: "Text goes here"
+  },
+  {
+    id: 3,
+    header: "Accordion Item #3",
+    subheader: "",
+    text: (
```
martypowell: this property name is deceiving, as the value here is html, not just text. Maybe `content` is a more appropriate name?

src/styles/partials/components/_accordion.scss
tmccoy529: done

src/styles/partials/components/_accordion-dark.scss
```diff
@@ -0,0 +1,32 @@
+.dg_section.dark {
```
martypowell, line 1: perfect

martypowell (dismissed):
a couple minor things.

src/components/site/Accordion.md
tmccoy529: changed to content

src/components/site/Accordion.js
tmccoy529: desrtructed this

src/components/site/Accordion.js
tmccoy529: Done

src/components/site/Accordion.js
tmccoy529: Meh, I think yours is fine. Its changed.

src/styles/partials/components/_accordion.scss
martypowell: @tmccoy529 still missing this one.

src/styles/partials/components/_accordion.scss
```diff
@@ -0,0 +1,97 @@
...
+      padding-bottom: $padding-small;
+      font-size: $h1-font-size;
+      @include icon-content($icon-arrow-down);
+      top: 0;
+      right: 0;
+    }
+  }
+}
+
+.collapse {
```
martypowell, line 98: these probably need to move out of here eventually but not blocking, as i don't think we have a good place for them.

src/components/site/Accordion.js
```diff
@@ -0,0 +1,74 @@
...
+
+const Accordion = props => {
+  const { canHaveMultiplePanelsOpen, className } = props;
+  const { collapseItems = [] } = props;
+
+  return (
+    <div
+      className={`dg_accordion ${
+        canHaveMultiplePanelsOpen === true ? "canHaveMultiplePanelsOpen" : ""
+      } ${className}`}
```
martypowell: this is a good case for the classnames package.

See [here](https://github.com/baltimorecounty/dotgov-components/blob/e90b2124ff91edd90e0d908732d035c5bfcce46e/src/components/basic/containers/Section.js#L7) for the implementation. 

In your case, for this, I think you can use this nifty trick.

```js
const cssClasses = classnames('dg_accordion', { 'canHaveMultiplePanelsOpen': canHaveMultiplePanelsOpen }, classname);
```

martypowell: This will prevent you from having the undefined class that is rendered now.

martypowell: https://www.npmjs.com/package/classnames

tmccoy529: this was originally added to handle the dg_section and dark. Accordion is no longer adding dg_section I've removed it entirely.

src/styles/partials/components/_accordion.scss
```diff
@@ -0,0 +1,97 @@
...
+      top: $padding-small;
+      right: 0;
+    }
+  }
+}
+
+.collapsed .dg_accordion-btn::after {
+  position: absolute;
+  cursor: pointer;
+  @include icon-content($icon-arrow-right);
```
martypowell, line 95: this will also require the font-size change.

tmccoy529: done

martypowell:
@tmccoy529 this is on hold pending some feedback from the design team.

tmccoy529:
@martypowell per design I have removed the duo functionality for accordions. This is ready for review.

src/components/site/Accordion.js
```diff
@@ -2,16 +2,14 @@ import React from "react";
 import PropTypes from "prop-types";
 
 const Accordion = props => {
+<<<<<<< HEAD
```
martypowell: @tmccoy529 looks like you have a good ole merge conflict here.

martypowell (approved):

PR #155, opened 2019-10-04
Homepage assembly bug#14957

theilman7:
This addresses a bug on the home page, referenced in Bug #14957. Refactored the scss to include parents for the pseudo class `:focus`. Not having the element attached to a parent, causes a focused effect over `div` elements as well, when clicked. Thinking the `:disabled` element could potentially be refactored as well, but opening that up for feedback/ discussion first.

### DIFF
--- a/src/styles/partials/base/_state.scss
+++ b/src/styles/partials/base/_state.scss
@@ -2,16 +2,69 @@ a,
 button,
 a i,
 button i {
-	@include default-transition(color, background, border);
+  @include default-transition(color, background, border);
 }
 
-:focus,
-.focused {
-	outline: 2px solid $blue-lightest;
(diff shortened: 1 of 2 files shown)


src/styles/partials/base/_state.scss
```diff
@@ -2,16 +2,69 @@ a,
...
+    }
+
+    &.focused {
+      outline: 2px solid $blue-lightest;
+      outline-offset: 3px;
+    }
+  }
+}
+
+button {
```
martypowell, line 32: i think this logic needs to be updated in `_button.scss` instead of here.

martypowell:
closing in favor of #165 @theilman7

PR #161, opened 2019-10-09
Background image most popular

tmccoy529:
Addresses TFS Issue 14956: Popular Services BG Image Missing. This might not be the correct way to handle this. But Ive added a prop for Icon Headings to allow them to take more classes then the one assigned to handle background images. Added styling to put a background image in a section. Added background to Icon Links on a darm background to keep the blue color even if they lay on top of a background image. Added reactive changes based on medium and smaller break points.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@baltimorecounty/dotgov-components",
-  "version": "0.1.36",
+  "version": "0.1.40",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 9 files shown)


martypowell:
@tmccoy529 icon heading pr is submitted for review, I'll update this thread when the npm package for icon headings has been updated.

martypowell:
talked with @tmccoy529 over chat, and we are going to try to attempt to replicate similar functionality as the page header

tmccoy529:
@martypowell this has been updated to use a prop for the background image instead of direct css.

martypowell:
@tmccoy529 this is just pending an image switch since, css will handle the color, not the image.

martypowell (approved):

PR #162, opened 2019-10-10
Update icon heading

martypowell:
As a result of #161 we discovered that the icon heading did not support sections with an image as a background.

This is an attempt to fix the styles to work regardless of a the background of the container element.

### DIFF
--- a/src/styles/partials/components/_icon-heading.scss
+++ b/src/styles/partials/components/_icon-heading.scss
@@ -18,8 +18,8 @@ $image-border-adjustment-margin: -$border-width-large * 3;
   margin-bottom: $default-margin;
 
   img {
-    max-width: 75px;
-    width: 100%;
+    height: 75px;
+    width: 75px;
   }
 
   &:before,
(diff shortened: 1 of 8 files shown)


src/styles/partials/components/_icon-heading.scss
```diff
@@ -6,91 +6,39 @@ $image-border-adjustment-margin: -$border-width-large * 3;
...
-    }
+  &:before {
+    margin: 10px $icon-spacing 10px 0;
   }
 
-  .dg_icon-heading__image-container {
-    margin-bottom: $default-margin - $image-border-adjustment-margin; /* account for border adjustment */
+  &:after {
+    margin: 10px 0 10px $icon-spacing;
   }
```
tmccoy529, line 42: isn't 10px just $margin-small

martypowell:
@tmccoy529 I made the changes you suggested in person, maybe you didn't hit submit on your review?

tmccoy529 (dismissed):

src/styles/partials/components/_icon-heading.scss
martypowell: good catch, updated all the variables in this file.

PR #164, opened 2019-10-11
Remove title from alternate list item

sgrg1:
Per Marty's review I am removing tile from alternateListItem for bug #14914


martypowell (dismissed):
@sgrg1 for the sake of time, i just removed the props from the component as well.

PR #167, opened 2019-10-15
Breadcrumb Tweaks

martypowell:
- Fixed spacing between links and icons
- Removed clickability on arrows
- Adjusted Icon to use standard arrow icon
- Fixed focus to be on line only

### DIFF
--- a/src/styles/partials/components/_breadcrumbs.scss
+++ b/src/styles/partials/components/_breadcrumbs.scss
@@ -23,6 +23,7 @@ $breadcrumb-arrow-size: 20px;
 
     &:last-child {
       margin: 0 $default-padding;
+      font-weight: $font-medium;
     }
   }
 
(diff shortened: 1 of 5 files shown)


src/styles/partials/components/_breadcrumbs.scss
```diff
@@ -1,3 +1,5 @@
+$breadcrumb-arrow-size: 20px;
+
```
tmccoy529, line 2: Should this be moved into a variables page? I feel like if its used enough to warrant a variable then maybe it needs to be available globally. But not blocking.

tmccoy529 (approved):

src/styles/partials/components/_breadcrumbs.scss
martypowell: not sure, its not a bad idea, but I think this is just specific to breadcrumbs right now, and doesn't match up with the default. Worth coming back to.

PR #169, opened 2019-10-15
Card container

theilman7:
Addresses User Story number [14767](https://oittfs.bcg.ad.bcgov.us/BAUCollection/BaltimoreCounty.Gov%20Website/_workitems/edit/14767). This creates a container for card content that can accept existing components and html elements. I also designed this with the ability to accommodate  a center and left justification on the content. In addition, the cards will inherit the height of the tallest card.


martypowell:
@theilman7 I suspect we want our cards to have to sub sections.

```html
<div class="dg_card">
  <div class="dg_card__content">
    <!-- stuff goes here -->
  </div>
  <div class="dg_card__footer">
   <!-- buttons go here -->
   </div>
</div>
```

This will allow us to use flex box to align the buttons.  Which would be something like.

```css
.dg_card {
  display: flex;
  flex-direction: column;
  justify-content: space-between;
}
```

I would also use the grid to layout the cards, instead of manual styles for now. Please let me know if you have any questions.

martypowell:
@theilman7 your branch is out of date as well.

src/styles/partials/components/_card-container.scss
```diff
@@ -0,0 +1,42 @@
...
+  position: relative;
+  margin-bottom: $margin-large;
+  padding: $padding-large;
+  width: 100%;
+
+  .dg_card-content-center {
+    display: block;
+    text-align: center;
+
+    .dg_icon-container {
```
martypowell: I don't think you need to dump this in here, I think you have it covered on line 36. It wouldn't be specific to centered content

theilman7: Updated and committed.

martypowell:
per our chat, @theilman7 is going to use the bootstrap utilities to help out with alignment inside the card, and will remove the alignment specific components. @theilman7 your branch is out of date, go ahead and get the latest please.

theilman7:
Committed discussed updates. Branch has been brought up-to-date, CSS was cleaned up based upon feedback and the .md file has been updated.

src/styles/partials/components/_card-container.scss
```diff
@@ -0,0 +1,26 @@
...
+  flex-direction: column;
+  justify-content: space-between;
+  position: relative;
+  margin-bottom: $margin-large;
+  padding: $padding-large;
+  width: 100%;
+
+  .dg_icon-container {
+    color: $gold;
+    font-size: 60px;
```
martypowell: i suspect maybe this should be an icon variable?

src/styles/partials/components/_card-container.scss
```diff
@@ -0,0 +1,26 @@
...
+    margin-bottom: 20px;
+  }
+
+  .dg_card__content {
```
martypowell: I think if this is the default style you would just need to add text-align: center to dg_card and remove lines 18-25?

src/styles/partials/components/_card-container.scss
theilman7: $largest-icon-size seemed a bit large and $large-icon-size seemed a bit small. If choosing between the two, I'd probably go with $largest-icon-size. The change has been committed to preview.

src/styles/partials/components/_card-container.scss
theilman7: Removed lines and committed.

martypowell (approved):

Issue #120, opened 2019-09-13
Step List - Static


danfox01:
## Component Name

StepList - Static

## Description

The step list is a visual representation of a process that's designed to make complex content more easy to scan and understand. It consists of large numbered steps, a bold label/heading, and a text description of each step.

## Example or Mockup

See the [design document](https://xd.adobe.com/spec/ad47a25b-c580-4389-756b-04b95b9d8a00-db76/screen/b678a40a-77d6-41bd-82f5-675a07caf62d/14-Step-Lists/).

#### Page Load

**Scenario**: Web services wants to display a static step list
**Given**: The user visits a page with a step list
**When**: The page loads
**Then**:
  - No buttons are visible in the step list
  - All Step's details are visible

**Scenario**: Web services wants to display all items as collapsed
**Given**: The user visits a page with a step list
**When**: The page loads
**Then**:
  - A Show All button is visible
  - All Step's details are hidden
  - Each Step includes a "Show" button

**Scenario**: Web services wants to display all items as expanded
**Given**: The user visits a page with a step list
**When**: The page loads
**Then**:
  - A Hide All button is visible
  - All Step's details are visible
  - Each Step includes a "Hide" button

#### Interactions

**Scenario**: A user wants view a collapsed section
**Given**:
  - The section details are hidden
  - The section has a "Show" button
  - The step list is NOT static

**When**: A user selects a section "Show" button
**Then**:
  - The "Show" button's text changes to "Hide"
  - The details are shown for that step

**Scenario**: A user wants hide a visible section
**Given**:
  - The section details are visible
  - The section has a "Hide" button
  - The step list is NOT static

**When**: A user selects a section "Hide" button
**Then**:
  - The "Hide" button's text changes to "Show"
  - The details are hidden for that step

**Scenario**: A user wants show all steps details
**Given**:
  - At least one section details section is hidden
  - At least one section button display "Show"
  - The step list is NOT static

**When**: A user selects then "Show All" button
**Then**:
  - The "Show All" button's text changes to "Hide All"
  - Each Step's detail section is visible

**Scenario**: A user wants hide all steps details
**Given**:
  - All section details must be visible
  - At section button display "Hide"
  - The step list is NOT static

**When**: A user selects then "Hide All" button
**Then**:
  - The "Hide All" button's text changes to "Show All"
  - Each Step's detail section is hidden

**Scenario**: The "Hide All" button should change to "Show All"
**Given**:
  - All section details must be visible
[...]

PR #177, opened 2019-10-18
Step List

martypowell:
Added #120 Step List

- Added styles
- Added examples for dynamic and static versions
- Added tests to verify props


src/styles/partials/components/_step-list-dark.scss
```diff
@@ -0,0 +1,35 @@
...
+
+    .dg_step-list__toggle-btn__btn-text,
+    .dg_step-list__show-all-btn {
+      display: none;
+    }
+
+    .dg_step-list__list > li::before {
+      background: $pale-blue;
+      border-color: $white;
+      color: $white;
```
martypowell, line 32: this can be removed.

PR #179, opened 2019-10-21
Fixing bugs

tmccoy529:
Addressing bugs #15093 #15094 and #15095 in TFS

Added h1-h3 in the accordion sass file to change color to $black from gray
Removed gold border and padding in dark for h1
Added div to image in md file to handle reactive images

### DIFF
--- a/src/components/site/Accordion.md
+++ b/src/components/site/Accordion.md
@@ -38,15 +38,13 @@ const collapseItems = [
     header: "Accordion Item #4",
     subheader: "",
     content: (
-      <img
-        width="300"
-        height="200"
-        alt="Stuff goes here"
-        src="//baltimorecountymd.gov/sebin/t/t/homepage-county-executive.jpg"
-        border="0"
-        vspace="0"
-        hspace="0"
-      />
+      <div id="dg_main-content">
+        <img
+          class="dg_image"
+          alt="Stuff goes here"
+          src="//baltimorecountymd.gov/sebin/t/t/homepage-county-executive.jpg"
+        />
+      </div>
     )
   },
   {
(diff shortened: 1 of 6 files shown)


src/components/site/Accordion.md
```diff
@@ -38,15 +38,13 @@ const collapseItems = [
...
-      <img
-        width="300"
-        height="200"
-        alt="Stuff goes here"
-        src="//baltimorecountymd.gov/sebin/t/t/homepage-county-executive.jpg"
-        border="0"
-        vspace="0"
-        hspace="0"
-      />
+      <div id="dg_main-content">
```
martypowell: Note that in this case, you are repeating the same id `dg_main-content` multiples times which is invalid. There can only be one id per page, this will generate multiples. I'm assuming you need the styles associated with 'dg_main content'?

src/styles/partials/components/_accordion-dark.scss
```diff
@@ -25,6 +25,10 @@
       a {
         color: $white;
       }
+      h2 {
```
martypowell: these styles aren't dark specific, so they should go into the _accordion.scss file. If for some reason this is an issue, please let me know and we can talk through this.

src/components/site/Accordion.md
tmccoy529: Shoot I though that was a class. I didnt even realize it was ID. Ugh yes, I'll have to pull those styles into this.

martypowell: per our chat, this id can be duplicated in the components until we come up for a solution for this problem.

src/components/site/Accordion.md
```diff
@@ -4,7 +4,7 @@ const collapseItems = [
     id: 1,
     header: "Accordion Item #1",
     subheader: "This is a subheader",
-    content: <h2>Test Heading Level 2</h2>
+    content: <h3>Test Heading Level 3</h3>
```
martypowell, line 7: this was adjusted to an h3 per design, as the accordion would never contain an h2, an h2 might be used to describe that instead

martypowell (approved):

PR #180, opened 2019-10-21
Added keys to child elements to clear console error

tmccoy529:
Fixed a console error inside of Summary List that was complaining about keys for child elements.

### DIFF
--- a/src/components/site/SummaryList.js
+++ b/src/components/site/SummaryList.js
@@ -7,13 +7,10 @@ const SummaryList = props => {
     <div className="dg_summary-list">
       {items.map(item => (
         <div key={item.id} className="dg_summary-list__item">
-          <i key={`summaryListItemIcon${item.id}`} className={item.icon}></i>
-          <div key={`summaryListItem${item.id}`}>
-            <h3 key={`summaryListItemHeader${item.id}`}>{item.heading} </h3>
-            <div
-              key={`summaryListHidden${item.id}`}
-              dangerouslySetInnerHTML={{ __html: item.content }}
-            />
+          <i className={item.icon}></i>
+          <div>
+            <h3>{item.heading} </h3>
+            <div dangerouslySetInnerHTML={{ __html: item.content }} />
           </div>
         </div>
       ))}
--- a/src/components/site/SummaryList.js
+++ b/src/components/site/SummaryList.js
@@ -1,15 +1,31 @@
 import React from "react";
 
 const SummaryList = props => {
-  const { items = [] } = props;
+  const { id, items = [] } = props;
   return (
-    <div className="dg_summary-list">
+    <div id={id} key={id} className="dg_summary-list">
       {items.map(item => (
-        <div key={item.key} className="dg_summary-list__item">
-          <i className={item.icon}></i>
-          <div>
-            <h3>{item.heading} </h3>
-            <div dangerouslySetInnerHTML={{ __html: item.content }} />
+        <div id={item.id} key={item.id} className="dg_summary-list__item">
+          <i
+            id={`summaryListItemIcon${item.id}`}
+            key={`summaryListItemIcon${item.id}`}
+            className={item.icon}
+          ></i>
+          <div
+            id={`summaryListItem${item.id}`}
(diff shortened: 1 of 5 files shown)


martypowell:
@tmccoy529 awesome, thank you.

src/components/site/SummaryList.js
```diff
@@ -1,20 +1,44 @@
 import React from "react";
+import PropTypes from "prop-types";
 
 const SummaryList = props => {
-  const { items = [] } = props;
+  const { id, items = [] } = props;
   return (
-    <div className="dg_summary-list">
+    <div id={id} key={id} className="dg_summary-list">
```
martypowell: i don't think the  `key` and `id` is need here, only inside the map?

src/components/site/SummaryList.js
```diff
@@ -1,20 +1,44 @@
 import React from "react";
+import PropTypes from "prop-types";
 
 const SummaryList = props => {
-  const { items = [] } = props;
+  const { id, items = [] } = props;
```
martypowell: in this case, i think you can completely remove id now. From this line and the prop-types. Will be completely optional, and probably won't have an id, but not sure.

src/components/site/SummaryList.js
```diff
@@ -1,20 +1,44 @@
...
     <div className="dg_summary-list">
       {items.map(item => (
-        <div key={item.key} className="dg_summary-list__item">
-          <i className={item.icon}></i>
-          <div>
-            <h3>{item.heading} </h3>
-            <div dangerouslySetInnerHTML={{ __html: item.content }} />
+        <div id={item.id} key={item.id} className="dg_summary-list__item">
+          <i
+            id={`summaryListItemIcon${item.id}`}
```
martypowell: just noticed id and key are everywhere, think you should only need this in the container div.

Something like is probably overkill

```jsx
<div
            id={`summaryListItem${item.id}`}
...
```

src/components/site/SummaryList.md
```diff
@@ -41,40 +41,48 @@ const publicWorksSummaryItems = [
   }
 ];
 
-<SummaryList items={publicWorksSummaryItems} />;
+<SummaryList id="summaryList-PublicWorks" items={publicWorksSummaryItems} />;
```
martypowell: we should probably remove this from the example, especially since the snippet does not have an id.

src/components/site/SummaryList.js
martypowell: also, not reflected in the markdown snippet, which is a compelling reason to remove.

martypowell (changes requested):
a couple of minor cleanup tasks

src/components/site/SummaryList.md
tmccoy529: removed

src/components/site/SummaryList.js
tmccoy529: id's removed

src/components/site/SummaryList.js
tmccoy529: removed id

src/components/site/SummaryList.js
tmccoy529: done

src/components/site/SummaryList.js
```diff
@@ -1,20 +1,29 @@
...
   return (
     <div className="dg_summary-list">
       {items.map(item => (
-        <div key={item.key} className="dg_summary-list__item">
-          <i className={item.icon}></i>
-          <div>
-            <h3>{item.heading} </h3>
-            <div dangerouslySetInnerHTML={{ __html: item.content }} />
+        <div key={item.id} className="dg_summary-list__item">
+          <i key={`summaryListItemIcon${item.id}`} className={item.icon}></i>
```
martypowell: sorry, key can also be removed, form the children, just needs to be on the parent div.

martypowell (approved):

martypowell:
merged on behalf of @tmccoy529. This will go in on the next component release, either modal or alerts, depending which comes first.

PR #181, opened 2019-10-22
W3 Modal Refactor

martypowell:
Worked with @sgrg1 to refactor the modal to use the [w3 modal](https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html). This will be in place of #160.

PR #185, opened 2019-10-23
Tiny icon link

martypowell:
For use with #181 

Adds a tiny version of the circular icon-link

### DIFF
--- a/src/components/basic/Button.md
+++ b/src/components/basic/Button.md
@@ -276,6 +276,36 @@ Html Snippet:
 </a>
 ```
 
+#### Tiny
+
+```jsx
+import IconLink from "./IconLink";
+<div
+  style={{ display: "flex", justifyContent: "space-between", padding: "10px" }}
+>
+  <IconLink
+    size="tiny"
+    href="#linkbutton-default"
+    id="linkbutton-default"
+    type="circle"
+    icon="far fa-times"
+    description="View your favorite default items."
+  />
+</div>;
+```
+
+Html Snippet:
+
+```html
+<a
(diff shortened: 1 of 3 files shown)


src/styles/partials/components/_link.scss
```diff
@@ -62,6 +63,14 @@ $small-icon-link-size: 50px;
     background: $gray-light;
   }
 
+  &.dg_icon-link--tiny {
+    padding: $padding-smallest $padding-small;
+    i {
+      @include icon-size(18px, 25px);
+      line-height: 27.5px;
```
martypowell, line 75: add a comment to explain why this is this way.

martypowell: This has been added.

src/components/basic/Button.md
```diff
@@ -276,6 +276,36 @@ Html Snippet:
 </a>
 ```
 
+#### Tiny
```
theilman7, line 279: Overall looks good. Do we want to add an example of this on a dark bg as well in the markdown file?

martypowell: The reason i didn't is I'm confident it won't display weird on dark backgrounds, since its just smaller?

theilman7: It does. Only reason I would add is just to balance out the default versions of the button section with the dark button section, since one shows three versions of the icon button and the dark version shows two. That's me just looking at the section as a single piece. I think functionally this is fine and not having it as a dark section in the markdown file is not blocking, so this should be good to go.

theilman7 (approved):
Updated markup looks good along with the local test of the component. Approved.

PR #188, opened 2019-10-28
Added Icon Button

martypowell:
There were cases where we need to use [IconLink](https://baltimorecounty.github.io/dotgov-components/#iconlink) as a button instead of an anchor element. This PR addresses that issue.


tmccoy529 (approved):
Looks good.

PR #189, opened 2019-10-28
Add margin card container bug#15173

sgrg1:
small fix, just added top margin.


martypowell:
@sgrg1 please take a look at the work item 15173 discussion, it has more details on addressing this issue. This only addresses cards, but this needs to account for other components. Please let me know if you have any question.

sgrg1:
@martypowell @tmccoy529 , added _content-top-bottom-margin.scss for top and bottom margin

martypowell:
> @martypowell @tmccoy529 , added _content-top-bottom-margin.scss for top and bottom margin

@sgrg1 this is still just addressing cards, let's talk about this one in person.

martypowell:
@sgrg1 I'm still not sure this will work. An example that breaks this is if we have an app where cards are the first content inside a container. There will be a huge gap between the top of the container and the start of the card list. Let's to put this one on hold until we have more time to figure out.

theilman7:
Would the spacing for this be a one size fits all approach, or would this be more suited to a case-by-case solution and solved with utilities? For example, Bootstrap uses utilities for [margin and padding utilities](https://getbootstrap.com/docs/4.3/utilities/spacing/) for such scenarios. We could of course setup our own version of the utilities to match our naming convention and functionality. Figured I would toss this out there for conversation.

martypowell:
closing, as we have made a decision to handle this at a component level for now.

PR #190, opened 2019-10-28
Modal tweaks

martypowell:
- Fix z-index on modals (ensure they are always on top)
- Fix button cursor for modal close button

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@baltimorecounty/dotgov-components",
-  "version": "0.1.64",
+  "version": "0.1.65",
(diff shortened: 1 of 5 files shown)


tmccoy529 (approved):
Looks fine to me

sgrg1:
@martypowell , looks good to me . I thought it was working fine with out a z-index too?

martypowell:
> @martypowell , looks good to me . I thought it was working fine with out a z-index too?

When i tested in the prototype is was not working with z-index, that's why i added. Realistically modals will always be on top of anything else, so in this case it ensures that is always true.

PR #191, opened 2019-10-28
fix responsiveness issue

sgrg1:
Fix responsiveness issue with alters Bug#15183

### DIFF
--- a/src/components/basic/Alert.js
+++ b/src/components/basic/Alert.js
@@ -20,14 +20,14 @@ const Alert = props => {
 
   return (
     <div className={cssClasses}>
-      <div className="dg_alert__container">
-        {icon && <i className={iconCssClasses}></i>}
-        <div className="dg_alert__content">
-          {type && isStatusAlert && (
-            <span className="dg_alert__status">{type}</span>
-          )}
-          {children}
-        </div>
+      <div className="dg_alert__content">
+        {type && isStatusAlert && (
+          <span className="dg_alert__status">
+            {icon && <i className={iconCssClasses}></i>}
+            {type}
+          </span>
+        )}
+        {children}
       </div>
     </div>
   );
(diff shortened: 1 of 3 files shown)


src/styles/partials/components/_alert.scss
```diff
@@ -3,17 +3,6 @@
   padding: $padding-large $default-padding;
```
martypowell, line 3: This rule needs to change a bit for icons https://github.com/baltimorecounty/dotgov-components/blob/5153da4d060050c991005d24fda31aa87483f756/src/styles/partials/components/_alert.scss#L37

font-size should be removed (it will inherit it from it's container) and the right padding should change from `20px` to `10px`

martypowell: another change in this file is on small screens and below the padding for `.dg_alert` should change to `10px` for the top and bottom instead of `20px` which is correct for screens above the small size.

martypowell (changes requested):
A couple of css updates.

src/styles/partials/components/_alert.scss
martypowell: fyi, @sgrg1 looks like your branch is out of date

sgrg1: @martypowell @tmccoy529 ,fixed issue with padding and merge with integration

martypowell (approved):
Looks good to me.

src/components/basic/Alert.js
```diff
@@ -20,14 +20,14 @@ const Alert = props => {
...
     <div className={cssClasses}>
-      <div className="dg_alert__container">
-        {icon && <i className={iconCssClasses}></i>}
-        <div className="dg_alert__content">
-          {type && isStatusAlert && (
-            <span className="dg_alert__status">{type}</span>
-          )}
-          {children}
-        </div>
+      <div className="dg_alert__content">
```
martypowell, line 23: @sgrg1 do we still need this div, I don't think we do.

PR #192, opened 2019-10-29
Alternate Table Layout

martypowell:
Addressed the following table layout

image

- Added examples to the style guide
- Adjusted styles to work on all tables

### DIFF
--- a/src/components/basic/table/Table.md
+++ b/src/components/basic/table/Table.md
@@ -37,7 +37,7 @@ import TableRow from "./TableRow";
 </Table>;
 ```
 
-If you aren't using react, you'll need to use the following html:
+Html Snippet:
 
 ```html
 <table class="table">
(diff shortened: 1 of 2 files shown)


tmccoy529 (dismissed):
Looks good to me

tmccoy529:
@martypowell I think this needs a merge with integration

sgrg1:
Looks good to me . Just have a question regarding border-collaspe:collaspe;  th{ backgrond-color:$white  and th{color: $black . Seems to me these property do not seems to be doing anything. Please correct me if I am wrong.

martypowell:
> Looks good to me . Just have a question regarding border-collaspe:collaspe; th{ backgrond-color:$white and th{color: $black . Seems to me these property do not seems to be doing anything. Please correct me if I am wrong.

It is not needed, our reset file handles this.

image

To see more about this check out in general, check out https://developer.mozilla.org/en-US/docs/Web/CSS/border-collapse#result

PR #194, opened 2019-10-29
SideBar Section Creation To Handle Bug 15135 in TFS

tmccoy529:
This is to address bug # 15135 Contact Us H2 Footer Spacing and Font Size in TFS.

Created a new component called SideBarSection. This will house a field to describe the side bar content that is reactive to screen size. At full screen it has normal top margin of 30px. When screen breaks at medium size a border is added to the top and a margin is added top and bottom of 30px. These are place holder margins until design gives us a definite on what they want. This also allows separation if additional side bar content is added with a new title.

Sidebar Section title font size has been adjusted to 20px for any h2 added also to stand apart from content. Again this is a place holder until design confirms or adjusts.


sgrg1:
@tmccoy529 @martypowell , looks good to me . I just what to know if we are still using "dg_sidebar" class because i donot see it on css.

tmccoy529:
@sgrg1 @martypowell this is actually a good question. The only css I saw is .sidebar I did not see .dg_sidebar however that class was used in the prototype so I was nervous removing it. So either .sidebar was supposed to be dg_sidebar or dg_sidebar shouldn't have been added to the prototype project. The .sidebar class is used throughout the components project though so I didn't think it was worth changing without a discussion.

src/components/site/SideBarSection.js
```diff
@@ -0,0 +1,19 @@
...
+    <div className="dg_sidebar sidebar">
+      <h2>{sideBarTitle}</h2>
+      <div>{children}</div>
+    </div>
+  );
+};
+
+SidebarSection.propTypes = {
+  /** The title to display */
+  sideBarTitle: PropTypes.string.isRequired
```
martypowell: Not blocking but since the component is name side, bar i think just `title` instead of `sideBarTitle` is probably sufficient here.

tmccoy529: makes more sense. updated

src/styles/partials/template/_sidebar.scss
```diff
@@ -16,9 +17,19 @@
...
+    font-size: $h3-font-size;
   }
 
   p {
     line-height: 30px;
   }
 }
+
+@media screen and (max-width: $medium-breakpoint) {
+  .sidebar {
```
martypowell: I think i maybe confused you here, the border should be at the top of each sidebar section regardless of the size of the screen.

martypowell: @tmccoy529 we alsoo probably want to create a class for this.

src/styles/partials/template/_sidebar.scss
```diff
@@ -1,4 +1,5 @@
 .sidebar {
```
martypowell: I'm not sure why this is like this (i probably did it) but i suspect this should be `dg_sidebar`

sgrg1 (dismissed):
looks good to me

src/styles/partials/template/_sidebar.scss
tmccoy529: updated to dg_sidebar

src/styles/partials/template/_sidebar.scss
tmccoy529: this section was actually reworked and has changed a bit.

src/components/foundation/BodyExample.js
```diff
@@ -1,39 +1,43 @@
-import React from 'react';
+import React from "react";
 
 const headingExampleStyles = {
-	borderTop: '1px solid lightgray',
-	padding: '20px'
+  borderTop: "1px solid lightgray",
```
martypowell, line 4: did anything apart from formatting change in this file?

martypowell (commented):
@tmccoy529 I think this looks good, we will need to sit down with @sdurphy and make sure the styles are right.

martypowell:
@tmccoy529 as mentioned before, looks good but putting on hold until @sdurphy is back.

src/components/foundation/BodyExample.js
tmccoy529: that page was using sidebar instead of dg_sidebar and that was all that was manually changed. Any other changes must be formatting.

src/components/site/SideBarSection.md
```diff
@@ -0,0 +1,75 @@
...
+  }
+];
+const sideBar = <SummaryList items={publicWorksSummaryItems} />;
+const followUsSideBar = <SummaryList items={followUsSummaryItems} />;
+
+<div>
+  <SidebarSection title="Contact Us">{sideBar}</SidebarSection>
+  <SidebarSection title="Follow Us">{followUsSideBar}</SidebarSection>
+</div>;
+```
```
theilman7, line 75: Do we need an HTML Snippet on this?

martypowell: > Do we need an HTML Snippet on this?

Yes, good catch.

tmccoy529: Added snippet

src/styles/partials/template/_sidebar.scss
```diff
@@ -1,4 +1,6 @@
-.sidebar {
+.dg_sidebar {
```
theilman7, line 1: In the pull request description, it mentions "At full screen it has normal top margin of 30px". If this is the case, I would add a `margin-top: 30px;`(using the corresponding variable of course) under the `dg_sidebar` class.

src/styles/partials/template/_sidebar.scss
```diff
@@ -1,4 +1,6 @@
...
+  border-top: $border-width-large $gray-light solid;
+  margin-bottom: $margin-larger;
```
theilman7, line 2: The PR also mentions "When screen breaks at medium size a border is added to the top..." Should the `border-top` property be moved under the media-query or am I interpreting this incorrectly?

src/styles/partials/template/_sidebar.scss
```diff
@@ -10,15 +12,23 @@
...
+    font-size: $h3-font-size;
   }
 
   p {
     line-height: 30px;
   }
 }
+
+@media screen and (max-width: $medium-breakpoint) {
+  .dg_sidebar:first-of-type {
```
theilman7, line 34: Wondering if the `first-of-type` selector is necessary. In theory, there should only be one dg_sidebar per page, correct? Thinking this should simply be .`dg_sidebar`.

src/styles/partials/template/_sidebar.scss
tmccoy529: So this is still sort of up for debate. As @martypowell explained it, the only time we would be adding the 30px top was when the screen was small enough to have it be single column wide. This was then put separation between the item above. So that is why it hasn't been added to that class.

src/styles/partials/template/_sidebar.scss
tmccoy529: We added this for in case of adding multiple side bars. Although this is probably true that only one side bar will exist per page the XD document did show a possibility of having more then one such as Contact Us and Follow Us.

src/styles/partials/template/_sidebar.scss
tmccoy529: The margin bottom is added for seperation between multiple side bars. Margin top is added to the media query to seperate it from any element above it once the screen shrinks to a smaller size.

src/components/site/SideBarSection.js
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+const SidebarSection = props => {
+  const { title, children } = props;
+  return (
+    <div className="dg_sidebar">
```
martypowell: I know this is a rename thing, but it probably makes sense to name teh class based on the componet. In this case

`dg_sidebar__section`

`dg_sidebar` sounds like it's container, which I think we will maybe need.

src/styles/partials/template/_sidebar.scss
```diff
@@ -10,15 +12,23 @@
...
     font-family: $default-heading-font;
     font-size: $smallest-heading-font-size;
+    margin: $default-margin 0;
   }
 
   h2 {
     border: none;
     letter-spacing: 10px;
     font-weight: $font-bold;
+    font-size: $h3-font-size;
```
martypowell: In our other system for work, @sdurphy suggested this should be 18px.

src/styles/partials/template/_sidebar.scss
martypowell: @tmccoy529 I thought we had talked about keeping it consistent with add margin to the bottom of things? Would be something like

```scss
.dg_sidebar {
  margin-bottom: $margin-larger;

  &:last-child {
    margin-bottom: 0;
  }
}
```

If you inspect any of the typography, you'll notice the top margin has been removed but a bottom margin exists. Probably something that I could have brought up for discussion. 

My argument is just to keep it consistent, if `top` is the way to go, then I'm fine, we should just switch the other items.

src/styles/partials/template/_sidebar.scss
martypowell: @tmccoy529 @theilman7 I think there may be some confusion. Right now, dg_sidebar is each of these:

image

The bug is actually talking about this space on smaller screens.

image


Tim is trying to fix both as part of this pr (I think)

src/styles/partials/template/_sidebar.scss
martypowell: @tmccoy529 after going through this pr from the bottom up, I realize what you are doing now. My comment may be valid and may not be depending on what happens with the other comments. I'll wait a bit until the other comments are addressed to chime back in on this particular conversation.

src/styles/partials/template/_sidebar.scss
martypowell: I guess the real question goes down to the naming and teh structure for that.

Is it simply just a number of dg_sidebars?

- dg_sidebar
- dg_sidebar
- dg_sidebar

Or is the structure

- dg_sidebar
  - dg_sidebar__section
  - dg_sidebar__section
  - dg_sidebar__section

src/components/site/SideBarSection.js
tmccoy529: Well not really. That div IS the entire sidebar with sections. Maybe we should wrap the h2 in a div with a .dg_sidebar__section class since that is an actual "section header"? Would then look something like this which is probably better. 

image

src/styles/partials/template/_sidebar.scss
tmccoy529: See below comment. I think this is correct having an additional section class below sidebar

src/components/site/SideBarSection.js
tmccoy529: This has been done

src/styles/partials/template/_sidebar.scss
tmccoy529: changed to 18

src/styles/partials/template/_sidebar.scss
tmccoy529: Per our discussion, dg_main_content needs to have a bottom margin added thereby avoiding the media query entirely for side bar but side bar also needs a bottom of 40 for stacking. This has been done.

src/components/site/SideBarSection.js
martypowell: thanks @tmccoy529 this makes a lot of sense to me.

martypowell (dismissed):
@tmccoy529 this looks great to me, @theilman7 do you have any other questions?

martypowell (approved):

theilman7 (commented):
Looks good to me.

PR #198, opened 2019-10-31
change the font size to 18px

sgrg1:
fix font issue

### DIFF
--- a/src/components/basic/Card.md
+++ b/src/components/basic/Card.md
@@ -13,89 +13,81 @@ import CardFooter from "./CardFooter";
 import Section from "./containers/Section";
 
 <Section>
-  <div className="row">
-    <div className="d-flex col-md-4">
-      <Card>
-        <CardContent>
-          <div class="dg_card__content">
+  <div className="container">
+    <div className="row">
+      <div className="d-flex col-md-4">
+        <Card>
+          <CardContent>
             <div className="dg_icon-container">
               <i className="fas fa-star" aria-hidden="true"></i>
             </div>
             <h3>Centered With Icon</h3>
             <p>.8 acres</p>
-            <p>
-              <a href="https://www.google.com/maps">
-                300 Lennox Avenue Room 202 Towson, Maryland 21286
-              </a>
-            </p>
-          </div>
-        </CardContent>
-        <CardFooter>
-          <button type="button" className="dg_button">
-            Learn More
-          </button>
-        </CardFooter>
-      </Card>
-    </div>
-    <div className="d-flex col-md-4">
-      <Card>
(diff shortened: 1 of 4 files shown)


theilman7 (dismissed):
Looks good. Font size for the `p` tag appears to be 18px.

src/components/basic/Card.md
```diff
@@ -22,11 +22,13 @@ import Section from "./containers/Section";
           </div>
           <h3>Centered With Icon</h3>
           <p>.8 acres</p>
-          <p>
-            <a href="https://www.google.com/maps">
-              300 Lennox Avenue Room 202 Towson, Maryland 21286
-            </a>
-          </p>
+          <div className="dg_card__content">
+            <p>
```
martypowell: @sgrg1 I think your changes are right, but your example doesn't make sense. In the html here, you are adding a `p` tag to control the style, but you shouldn't be adding html to adjust the style, thats why we have css. So in this case, you probably want to take the rule you created and add `a` to it

```scss
.dg_card__content {
  p,
  a {
    font-size: $small-heading-font-size;
  }
}
```

And then update your examples by removing the `p` tag you added. You will want to test and make sure this css rule doesn't impact any of the other markup within the container.

martypowell (changes requested):
See feedback about the example.

src/components/basic/Card.md
theilman7: Worked with @sgrg1 to make some updates. @martypowell May want to take a second look.

martypowell: we still have extra markup controlling style, we do not want to add extra tags when we don't need to, this keeps the site smaller which makes it quicker it also makes it easier to maintain.

A paragraph around an a makes sense in this scenario...

```html
<p>Find out more about <a href="/mr-fluffy">my dog</a>.</p>
```

However when the links is like it is here:

```html
 <a href="https://www.google.com/maps">
  300 Lennox Avenue Room 202 Towson, Maryland 21286
</a>
```
we don't need it.

We can handle that with css, so the rule I think @sdurphy is that links and paragraph text inside a card are $small-heading-font-size. The way to accopmlish that is using the css I listed above.

**So items to resolve this:**

- Remove `<p>` tag around `<a>` in this file
- Adjust the css to make both p and a tags $small-heading-font-size within cards
- Test changes to make sure they don't blow anything else up in the example.

martypowell: if you have questions, please let me know.

sgrg1: work with travis to fix this issue.

martypowell (approved):

PR #199, opened 2019-10-31
removed focus outline for static step lists

tmccoy529:
Takes care of 15220 Static Step List should not allow for focus in TFS

Removes the focus around a static step list if tabbed or clicked.

### DIFF
--- a/src/styles/partials/components/_step-list-dark.scss
+++ b/src/styles/partials/components/_step-list-dark.scss
@@ -23,9 +23,6 @@
   .dg_step-list.static {
     .dg_step-list__toggle-btn {
       cursor: default;
-      &:focus {
-        outline: none;
-      }
     }
 
     .dg_step-list__toggle-btn__title {
--- a/src/styles/partials/components/_step-list-dark.scss
+++ b/src/styles/partials/components/_step-list-dark.scss
@@ -23,10 +23,13 @@
   .dg_step-list.static {
     .dg_step-list__toggle-btn {
       cursor: default;
+      &:focus {
+        outline: none;
+      }
     }
 
     .dg_step-list__toggle-btn__title {
-        color: $white;
+      color: $white;
     }
 
     .dg_step-list__toggle-btn__btn-text,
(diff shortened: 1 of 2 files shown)


src/styles/partials/components/_step-list-dark.scss
```diff
@@ -23,10 +23,13 @@
   .dg_step-list.static {
     .dg_step-list__toggle-btn {
       cursor: default;
+      &:focus {
```
theilman7: Not sure if we need this in _step-list-dark.scss as well. When testing locally and only using the outline: none; in _step-list.scss, this seems to still work as expected. I'm testing on mac so my experience could differ though.

src/styles/partials/components/_step-list.scss
```diff
@@ -106,6 +106,9 @@
 .dg_step-list.static {
   .dg_step-list__toggle-btn {
     cursor: default;
```
theilman7, line 108: Question for the group. Since the user is still able to open and close each item in the static step list, should the `cursor` property be set to `pointer` to communicate that functionality to the user?

sgrg1 (dismissed):
works on chrome. looks good to me

src/styles/partials/components/_step-list-dark.scss
martypowell: > Not sure if we need this in _step-list-dark.scss as well. When testing locally and only using the outline: none; in _step-list.scss, this seems to still work as expected. I'm testing on mac so my experience could differ though.

@theilman7 this sounds right to me, typically only color changes are needed in the "theme" file, although i could see this potentially possible. I have not tested though.

src/styles/partials/components/_step-list.scss
martypowell: actually @theilman7 the static step list does not allow the user to interact. All steps are expanded and no buttons are visible.

martypowell (commented):
I think this looks good, but I'd take a look at @theilman7 comment, I think this is valid

martypowell (approved):

src/styles/partials/components/_step-list-dark.scss
tmccoy529: @theilman7 @martypowell confirmed this is was needed. Good catch Travis

tmccoy529: Not needed

PR #200, opened 2019-10-31
Ie11 overlaps svg#15221

sgrg1:
issue fixed


src/styles/partials/components/_card-container.scss
```diff
@@ -9,10 +9,14 @@
...
-
-  .dg_icon-container {
-    color: $gold;
-    font-size: $largest-icon-size;
-    margin-bottom: 20px;
+}
+.dg_icon-container {
+  color: $gold;
+  font-size: $largest-icon-size;
+  margin-bottom: 20px;
```
martypowell: this should use a margin variable.

src/styles/partials/components/_card-container.scss
```diff
@@ -9,10 +9,14 @@
...
+}
+.dg_card__content {
```
martypowell: was this added, if so why? I didn't see anything about this in the bug?

sgrg1: I believe it was from previous PR fix "change the font size to 18px". I must have created branch from that fix.

src/components/basic/Card.md
```diff
@@ -47,15 +49,17 @@ import Section from "./containers/Section";
             role="img"
             aria-label="Placeholder: Thumbnail"
           >
-            <rect width="100%" height="200px" fill="#C2C2C2"></rect>
+            <rect width="100%" height="150px" fill="#C2C2C2"></rect>
```
martypowell: it looks like this placeholder svg is being repeated over and over. This sounds like a good instance to create a small component within this file and then reference that.

Example:

```jsx
const ImagePlaceholder = props => <svg><rect ></rect></svg>;
```

Then you can use it in the cards, like so

```jsx
<CardContent>
  <ImagePlaceHolder />
...
```



Then in the future, if we have to tweak the svg, we only need to do it in one place.

src/styles/partials/components/_card-container.scss
martypowell: no problem, i was hitting these backwards, we can wait until that one is completed

src/styles/partials/components/_card-container.scss
sgrg1: done

src/components/basic/Card.md
sgrg1: done

martypowell (approved):

PR #202, opened 2019-11-01
moved section outside the container

martypowell:
Ensure the component sidebar section can stand alone without it's container for general styling.


tmccoy529 (dismissed):
Looks good

PR #205, opened 2019-11-04
Site Navigation

martypowell:
Adds Site Navigation to the Site Header

- Adds Hamburger Menu button and Slide out Navigation
- Fixes Spacing of Site Header is association with the hamburger and menu buttons


src/components/site/SiteNavigation.md
```diff
@@ -0,0 +1,25 @@
+```jsx
```
martypowell: This is just showing up with all of it's styles in the design system, so I think we should remove for now.

image

martypowell: especially since it's a single use component.

martypowell: removed.

src/components/site/SiteNavigation.js
```diff
@@ -0,0 +1,41 @@
...
+    <nav id="bc_site-nav" className="site-nav" role="navigation">
+      <SiteNavButton />
+      <div className="bc_site-nav__list-container invisible">
+        <div className="text-center">
+          <a href="/">
+            <CountySeal className="bc_site-nav__seal" />
+          </a>
+        </div>
+        <ul id="bc_site-nav__links" className="bc_site-nav__list">
+          {links.map(link => {
```
martypowell: I'd argue this coudl be simplified above, quick change, so i'll do it.

theilman7:
Taking a look now. Will reply with any comments, questions or feedback shortly.

theilman7:
Do we need a .test.js file for this component?

src/styles/partials/components/_icon-heading.scss
```diff
@@ -51,3 +51,9 @@ $icon-size: 30px;
   line-height: 29px; /* Manually calculated for this element in the browser */
   padding: 0 !important; /* overrides default h2 in a dark section */
 }
+
+@media screen and (max-width: $small-breakpoint) {
+  .dg_icon-heading {
+    letter-spacing: $base-spacing-unit * 2;
```
martypowell, line 58: @theilman7 @tmccoy529 @sgrg1 this was done because on small screens the text was causing overflow, in turn causing a horizontal scroll which we dont want...would a comment be helpful here?

src/styles/partials/components/_site-header.scss
```diff
@@ -34,11 +34,78 @@
...
+}
+
+@media screen and (max-width: $header-breakpoint) {
+  .dg_site-header .dg_brand {
+    .dg_brand-text {
+      border-right: none;
+      text-align: center;
+
+      position: absolute;
+      top: 12.5px;
```
martypowell: you should add a comemt here.

src/styles/partials/components/_site-header.scss
```diff
@@ -34,11 +34,78 @@
...
+      right: 0;
+      padding: 0;
+    }
+  }
+}
+
+$small-mobile-button-size: 65px;
+
+@media screen and (max-width: $small-breakpoint) {
+  .bc_site-nav__toggle-button,
```
martypowell, line 79: a comment will probably be helpful here as well.

src/styles/partials/components/_site-header.scss
```diff
@@ -34,11 +34,78 @@
...
+  .dg_mobile-search-button,
+  #fake-site-nav-button {
+    height: $small-mobile-button-size;
+    width: $small-mobile-button-size;
+    padding: 0;
+  }
+
+  .bc_site-nav__toggle-button {
+    .bc_site-nav__toggle-button__text {
+      font-size: 10px;
```
martypowell: can this be a variable, I think so, probably "tiny"-font-size

src/styles/partials/components/_site-header.scss
```diff
@@ -34,11 +34,78 @@
...
+    }
+
+    i {
+      font-size: 30px;
```
martypowell: check out the icon variables for this size, although i think this is a one off.

src/styles/partials/components/_site-header.scss
```diff
@@ -34,11 +34,78 @@
...
+  .dg_site-header {
+    .dg_brand {
+      .dg_brand-text {
+        top: 3px; /* visually calculated in the browser. */
+
+        span:first-of-type {
+          font-size: $smallest-font-size;
+        }
+        span:last-of-type {
+          font-size: 14px;
```
martypowell: should use variable here.

src/styles/partials/components/_site-nav.scss
```diff
@@ -0,0 +1,162 @@
...
+  .bc_site-nav__toggle-button {
+    position: absolute;
+    top: 10px; /* mimic the top border of the site */
+    transform: translateX(0);
+    z-index: $max-z-index;
+    @include default-transition(transform);
+  }
+
+  .bc_site-nav__seal {
+    max-width: 75px;
```
martypowell, line 38: do we have this seal stuff somewhere else? maybe footer?

src/styles/partials/components/_site-nav.scss
```diff
@@ -0,0 +1,162 @@
...
+  }
+}
+
+#bc_site-nav {
+  position: relative;
+  top: 0;
+
+  .bc_site-nav__toggle-button {
+    position: absolute;
+    top: 10px; /* mimic the top border of the site */
```
martypowell: we could probably set this border in a variable and use it.

src/styles/partials/components/_site-nav.scss
```diff
@@ -0,0 +1,162 @@
...
+        margin: $default-margin auto;
+        padding: $padding-large $default-padding;
+        position: relative;
+        text-decoration: none;
+        text-transform: uppercase;
+
+        i {
+          text-align: center;
+          margin-right: $margin-small;
+          width: 40px;
```
martypowell, line 76: I think this is similar to some other icon centering, maybe it can be a mixin. @theilman7 I think you worked on something like this, do you remember where it was?

src/styles/partials/components/_site-nav.scss
```diff
@@ -0,0 +1,162 @@
...
+}
+
+.is-active {
+  /* mimic the top body border */
+  &:after {
+    position: absolute;
+    content: "";
+    left: 0;
+    right: 0;
+    height: 10px;
```
martypowell: can you the created variable for border for site.

src/components/site/SiteNavButton.md
```diff
@@ -0,0 +1,3 @@
+```jsx
+<SiteNavButton />
```
theilman7: Only seeing a markdown file for the SiteNavButton. Will probably need a file for this if we want to test on the styleguide page as well.

src/styles/partials/components/_site-nav.scss
```diff
@@ -0,0 +1,162 @@
...
+  margin: auto;
+  transform: translateX(0);
+  @include default-transition(transform);
+
+  &:after {
+    content: "";
+    opacity: 0;
+    background: #000000;
+    @include default-transition(opacity);
+    // transition:  0.2s;
```
martypowell: please remove this bad boy.

src/styles/partials/components/_site-nav.scss
```diff
@@ -0,0 +1,162 @@
...
+      opacity: $default-overlay-opacity;
+      visibility: visible;
+    }
+  }
+}
+body.is-disabled {
+  overflow: hidden;
+}
+
+@include nav-menu-width($desktop-menu-width);
```
martypowell: probably worth addign a comment here.

src/styles/partials/components/_site-nav.scss
```diff
@@ -0,0 +1,162 @@
...
+
+/* breakpoint is specific to the site navigation only */
```
martypowell: this comment is not true here, this is a standard breaskpoint.

src/styles/partials/components/_site-nav.scss
```diff
@@ -0,0 +1,162 @@
...
+#bc_page {
+  position: relative;
+  margin: auto;
+  transform: translateX(0);
+  @include default-transition(transform);
+
+  &:after {
+    content: "";
+    opacity: 0;
+    background: #000000;
```
martypowell: this should be $black.

src/components/site/SiteNavButton.md
```diff
@@ -0,0 +1,3 @@
...
+```
```
theilman7: An HTML snippet could be helpful here, especially with the impending move to SE of these components.

src/components/site/SiteNavigation.js
```diff
@@ -0,0 +1,46 @@
...
+  return (
+    <nav id="bc_site-nav" className="site-nav" role="navigation">
+      <SiteNavButton />
+      <div className="bc_site-nav__list-container invisible">
+        <div className="text-center">
+          <a href="/">
+            <CountySeal className="bc_site-nav__seal" />
+          </a>
+        </div>
+        <ul id="bc_site-nav__links" className="bc_site-nav__list">
```
sgrg1, line 31: are we still using className="bc_site_nav__list"? I did not find this className in our css file.

src/components/site/SiteHeader.js
```diff
@@ -8,13 +8,16 @@ const SiteHeader = props => {
   const { status, heading, subHeading } = props;
   return (
     <header className="dg_header dg_site-header">
-      {/* Hamburger Menu Will Go Here */}
       <div className="dg_brand">
+        <div id="fake-site-nav-button"></div>
```
theilman7, line 14: Comparing the functionality to that within the prototype, feel like it may make more sense to include the button portion of SiteNavButton within the `fake-site-nav-button` div (if this is the case, I would recommend changing the naming of the class). I feel like having these elements detached may cause confusion in later builds, should someone new need to make a change down the road.

src/styles/partials/components/_icon-heading.scss
tmccoy529: It definitely wouldn't hurt.

src/components/site/SiteHeader.js
martypowell: @theilman7 I am not sure I am following. It sounds like you are suggesting that the hamburger button be placed inside of the fake-site-nav div?

src/components/site/SiteNavigation.js
martypowell: @sgrg1 I will check it out.

src/components/site/SiteHeader.js
martypowell: @theilman7 I added a comment, please let me know if that clarifies anything for you.

src/styles/partials/components/_icon-heading.scss
martypowell: added.

src/styles/partials/components/_site-header.scss
martypowell: done

src/styles/partials/components/_site-header.scss
martypowell: this is a one off, left for now with comment.

src/styles/partials/components/_site-header.scss
martypowell: changed to use variable.

src/styles/partials/components/_site-nav.scss
martypowell: added.

src/styles/partials/components/_site-nav.scss
martypowell: changed to variable.

martypowell:
> Do we need a .test.js file for this component?

@theilman7 I am not seeing what file you are talking about. I do think most of these site specific components will actually be removed from the style guide at some point or refactored to be more generic.

src/components/site/SiteNavigation.js
martypowell: changed scss to use this class.

src/components/site/SiteNavButton.md
martypowell: @theilman7 I removed the example (md file) in favor of adding a test. We shouldn't need this component in the documentation (at least I don't think) because we only have one instance of component and I dont think well need it again.

martypowell:
@baltimorecounty/front-end-developers this is ready for another look

src/components/site/SiteHeader.js
theilman7: Yes, I was thinking about having the hamburger button inside the fake-site-nav, but if its out of there for accessibility reasons, that should be fine.

sgrg1 (dismissed):
looks good to me

theilman7:
Noticed a bug when viewing on the [prototype](https://baltimorecounty.github.io/dotgov-prototype-2.0/#/).

When navigating to pages via the slide-out navigation, page opens with the body class `is-disabled`. This prevents the user from scrolling down the page.

martypowell:
@theilman7 @tmccoy529 can you guys take one more look please, this is a big one, and want to make sure we got it close.

src/styles/variables/_z-index.scss
```diff
@@ -1,3 +1,4 @@
 /* Used to set the item at the highest z-index and then directly below */
```
theilman7, line 1: While looking at other components, noticed there are other "fugitive" z-index variables of 999 inhabiting other classes out there. Just a word of caution for us all as we add more instances of this to be conscious of how these will interact with each other in relation to their [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context). This is more of a comment/ word of caution going forward to avoid adverse stacking scenarios.

martypowell: @theilman7 I think this is a good call, we should move these to variables.

src/styles/partials/components/_site-nav.scss
```diff
@@ -0,0 +1,161 @@
+/* Handles the widths and horizontal positioning for the slide-out menu and hamburger menu for desktop */
+@mixin set-navigation-width($width) {
+  #bc_site-nav .bc_site-nav__list-container {
+    width: $width;
+
+    &.invisible {
```
theilman7, line 6: More of question/ opening of discussion. Any reason using the invisible class vs selecting the parent div with `is-active`? Thinking this would accomplish the same functionality but with less markup, removing the need for the `invisible` class. In that case, I would imagine (speculating) the functionality of `.invisible` could be baked in to  `#bc_site-nav .bc_site-nav__list-container`. Thoughts?

martypowell:
per @theilman7 his concern is only a concern in the prototype, which we cna deal with for now.

PR #206, opened 2019-11-04
Added static step list item for demo purposes.

tmccoy529:
Added new static step list item component to handle the focusing issue in Bug 15220: Static Step List should not allow for focus.

New example has been added to component project to demonstrate the solution to remove buttons from static lists and make them spans.


src/components/basic/StepList.md
```diff
@@ -71,30 +80,30 @@ import StepListItem from "./StepListItem";
...
+          <StepListItemStatic
             id="step-3"
             buttonText="Step 3: Gather Required Documentation"
           >
             <p>Some really helpful content for step 3 will go here.</p>
-          </StepListItem>
+          </StepListItemStatic>
         </StepList>
       </div>
     </div>
```
theilman7, line 100: May want to add an HTML snippet example for these as well.

tmccoy529: Added snippet for these two types.

src/components/basic/StepListItemStatic.js
```diff
@@ -0,0 +1,29 @@
...
+      </div>
+    </li>
+  );
+};
+
+StepListItemStatic.propTypes = {
+  /** Unique id to identify each step, must be unique on the page */
+  id: PropTypes.string.isRequired,
+  /** Button Text to be displayed for the step, should identify which step it is. Example Step 1 - Do something really cool. */
+  buttonText: PropTypes.string.isRequired
```
martypowell: I know this is probably dumb, but int his case, this could be deceiving since we removed the button?

martypowell: fyi @tmccoy529

martypowell (approved):

src/components/basic/StepListItemStatic.js
```diff
@@ -0,0 +1,27 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+const StepListItemStatic = props => {
+  const { id, buttonText, children } = props;
+  return (
+    <li className="dg_step-list__list-section" id={`step-${id}`}>
+      <span className="dg_button-link dg_step-list__toggle-btn">
+        <span className="dg_step-list__toggle-btn__title">{buttonText}</span>
```
martypowell: sorry @tmccoy529 I was probably too brief in my previous comment, we still need this prop, its probably should be named something like `stepTitle` or something along those lines. If not we could have a missing title here.

martypowell (changes requested):
Minor prop thing, otherwise this looks good.

martypowell:
@tmccoy529 did we ever find out if this is still an issue?

theilman7:
Bringing this branch back up to date with integration. @martypowell @tmccoy529

theilman7:
Working on resolving this. Looks like the static step list is till expanding when clicked. Two solutions I can think of after peeling back the layers on this. So the root of the cause is the js is targeting the class `dg_step-list__details`. Upon doing so, it writes `display: none;` on that particular div. 

1. Update the Javascript so it targets by ID instead.
2. Update the class attached to the div within the static step-list, list items. I would imagine this would look something like `dg_static-step-list__details`.

Let me know which method you would prefer.

theilman7:
@martypowell 
1. Brought branch up to date.
2. Brought version number up to date.
3. Updated md file to represent the html code accurately.
4. Assuming this will not fully work until the js from the corresponding pr in the javascript repo has been uploaded.

src/components/basic/StepList.md
```diff
@@ -71,33 +71,142 @@ import StepListItem from "./StepListItem";
...
+      </div>
+    </li>
+  </ol>
+</div>
+```
+
+### Static Step List
+
+```html
+<ol class="dg_step-list__list">
```
martypowell: @theilman7 looks like maybe this is missing the "static" part of the markup?

theilman7: Added the wrapper to the html example in the md file. Pushed up in latest commit.

theilman7: @martypowell

martypowell:
@theilman7 this looks good, will merge on your behalf.

PR #207, opened 2019-11-19
Breadcrumbs se update

theilman7:
This addresses User Story 15299. Refactored SCSS and markup so the breadcrumbs work with the module provided by SiteExecutive. In addition, I have updated the .md file so the changes are reflected there as well.


theilman7:
The addition of `margin: 0px;` removes the excessive margin of the h1 tag in the page header, when used in SE. In the original comp, the title of the page is shown in a span but when used as an h1, it picks up a large margin on the bottom inside of the page header. See the text: "Prototyping Homepage" in the following [link](https://beta.baltimorecountymd.gov/prototyping/index.html).

src/styles/variables/_icons.scss
```diff
@@ -7,4 +7,5 @@ $smallest-icon-size: 0.5em;
 /* uses far fa-chevron- */
 $icon-arrow-down: "\f078";
 $icon-arrow-right: "\f054";
-$icon-arrow-default-size: 24px;
\ No newline at end of file
+$icon-home: "\f015";
```
martypowell, line 10: nice, thank you for doing this.

src/styles/partials/components/_breadcrumbs-dark.scss
```diff
@@ -1,19 +1,24 @@
...
-      a {
+      &:hover {
+        color: $blue-lightest;
+      }
+
+      &:after {
         color: $white;
+      }
+
+      &.breadcrumb-last {
```
martypowell: would :last-child work here instead?

src/components/site/PageHeader.md
```diff
@@ -28,44 +28,55 @@ const breadCrumbs = () => (
...
+    class="dg_page-header__backgroundImage"
+    src="http://baltimorecountymd.gov/sebin/j/x/public-works-header.jpg"
+    aria-hidden="true"
+  />
+  <div class="dg_page-header__container">
+    <div class="container">
+      <div class="row">
+        <div class="col">
+          <ul class="dg_breadcrumbs">
+            <li class="dg_breadcrumb">
```
martypowell: I think this maybe needs updated? I don't think `li` exists in the breadcrumbs anymore?

src/components/site/Breadcrumbs.md
```diff
@@ -2,16 +2,17 @@
...
 <Section>
   <Breadcrumbs>
-    <BreadcrumbLinkItem text="Departments" link="/departments" />
+    <BreadcrumbLinkItem text="Departments" link="/departments/health" />
     <BreadcrumbLinkItem
       text="Department of Health"
       link="/departments/health"
     />
-    <BreadcrumbItem text="Adoptable Pets" />
+    <BreadcrumbLastItem text="Adoptable Pets" link="/departments/health" />
```
martypowell: Would prefer to keep this `BreadcrumbLinkItem` if possible.

src/components/site/BreadCrumbLinkItem.js
```diff
@@ -4,9 +4,9 @@ import PropTypes from "prop-types";
 const BreadcrumbLinkItem = props => {
   const { link, text } = props;
   return (
-    <li className="dg_breadcrumb">
-      <a href={link}>{text}</a>
-    </li>
+    <a className="breadcrumb" href={link}>
```
martypowell: you should add proptypes to this component.

src/components/site/BreadCrumbLastItem.js
```diff
@@ -0,0 +1,20 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+const BreadcrumbLastItem = props => {
+  const { link, text } = props;
+  return (
+    <a className="breadcrumb breadcrumb-last" href={link}>
```
martypowell: as mentioned somewhere below i would like to try and avoid this component if possible. The only difference between this is a class, you can use the [classnames package](https://github.com/JedWatson/classnames) inside of `BreadcrumbLinkItem` to combine classes.  See [buttons](https://github.com/baltimorecounty/dotgov-components/blob/7f7d7733d3b03f8c567253603af62bb432f39c99/src/components/basic/Button.js#L7) for an example.

Let me know if you have questions about this.

martypowell (changes requested):
a couple minor comments and questions.

src/components/site/PageHeader.md
theilman7: That is correct. I overlooked the example in my original push. I'll update and re-commit.

theilman7: Update has been pushed with the most recent commit.

src/components/site/BreadCrumbLastItem.js
theilman7: Makes sense and seems like the way to go. Removed this component in the most recent commit. Carrying the conversation back over to the BreadCrumbLinkItem.js component, where I have refactored.

src/components/site/BreadCrumbLinkItem.js
theilman7: Refactored this item to accept classNames, so the required `.breadcrumb-last` class that SE creates can be added to the example in the components library. Let me know if this looks ok. I also updated other files that had dependencies for the BreadcrumbLastItem.js and BreadcrumbItem.js for that matter, since they are not necessary when using the refactored version of this component.

src/components/site/Breadcrumbs.md
theilman7: As mentioned in my last comment, updated to the new version of `BreadcrumbLinkItem`.

src/styles/partials/components/_breadcrumbs-dark.scss
theilman7: I think they would do the same thing, so I'm fine with either option since I'm not seeing many negatives on either. I will say, SE will always add the class `.breadcrumb-last` to the last child within the breadcrumbs module, so I think targeting `.breadcrumb-last` is catering more towards SE's intended usage of the styling. Let me know which method you would prefer, with that in mind. If we do end up using the `.breadcrumb-last` class as a parent, looking at it again, I should probably move the margin style under that as well.

src/components/site/BreadCrumbLinkItem.js
```diff
@@ -1,12 +1,14 @@
 import React from "react";
 import PropTypes from "prop-types";
+import classnames from "classnames";
 
 const BreadcrumbLinkItem = props => {
-  const { link, text } = props;
+  const { link, text, className } = props;
+  const BreadcrumbClasses = classnames("breadcrumb", className);
```
martypowell, line 7: @theilman7 "breadcrumb" assigned by se?

martypowell (dismissed):
Approved but please adjust to use last-child for last breadcrumb item

src/components/site/BreadCrumbLinkItem.js
theilman7: Yep, `.breadcrumb` is a class assigned to all children by SE.

src/styles/partials/components/_breadcrumbs-dark.scss
theilman7: Migrated all children of `.breadcrumb-last` to `:last-child`.

theilman7: Applied scss changes to _breadcrumbs-dark.scss as well.

martypowell:
@theilman7 I'm assuming you need me to push the magic button on this?

theilman7:
@martypowell Indeed. If all looks good, go right ahead. The only outstanding thing is Better Code Hub has an X next to it... but I'm not sure if the module is just acting up or if its something worth checking out. I tried to check Details, but it just took me to an intermediate page of some kind with a button.

martypowell:
> @martypowell Indeed. If all looks good, go right ahead. The only outstanding thing is Better Code Hub has an X next to it... but I'm not sure if the module is just acting up or if its something worth checking out. I tried to check Details, but it just took me to an intermediate page of some kind with a button.

@theilman7 this was nothing actually it's complaining that our code block is too long in the index file, but its really not, we are just exporting stuff here.

image

I guess maybe you could argue we could have an export for each folder, but for now i think this is fine.

martypowell:
@theilman7 merged

PR #210, opened 2019-12-13
Removed stray :after. Added :focus pseudo element.

theilman7:
@martypowell  @sdurphy Addresses bug "Breadcrumb Focus Outline Browser Differences" 

Two minor changes, one of which is more of a "housekeeping" item.

1. Added :`focus` pseudo-element to the `.breadcrumb `class, with the property `inline-block`. This allows the border to wrap around the link text in an even square, eliminating the odd shape observed previously, on tab focus.

2. Removed the orphan `:after` pseudo-element on line 17.


martypowell (approved):

martypowell:
@theilman7 sorry i missed this one. Merging now.

PR #212, opened 2020-01-02
Subpage tabs

theilman7:
Refer to User Story [15102](https://oittfs.bcg.ad.bcgov.us/BAUCollection/BaltimoreCounty.Gov%20Website/_workitems/edit/15102). Created subpage tabs component. This has a dependency on a corresponding javascript file to support the basic functionality, so I'm going to assume that should be added to the dotgov-javascript repo. Once that has been added, this item can be reviewed completely. Markup for SCSS and structure of react elements can be reviewed however.


src/components/site/SubPageTabItem.js
```diff
@@ -0,0 +1,22 @@
...
+      <a href={link}>{text}</a>
+    </li>
+  );
+};
+
+SubPageTabItem.propTypes = {
+  /** The copy that exists inside the a tag */
+  text: PropTypes.string.isRequired,
+  /** The link or href for the a tag */
+  link: PropTypes.string
```
martypowell: is link required?

src/components/site/SubpageTabs.md
```diff
@@ -0,0 +1,23 @@
...
+    <SubPageTabItem text="Media Contact Information" link="/home" />
+  </SubPageTabs>
+</Section>;
+```
+
+Html Snippet:
+
+```html
+<div class="dark">
+  <ul></ul>
```
martypowell: example looks likes its missing some markup.

martypowell (changes requested):
A couple questions, and small changes.

This is something that is giong to have to work wtih vanilla js right?

martypowell:
@theilman7 looks like you are out of date as well.

src/components/site/SubPageTabItem.js
theilman7: You are correct. Updated and re-committed.

src/components/site/SubpageTabs.md
theilman7: Setup the .md file while I was fine-tuning the markup and completely forgot to circle back and update this. Added relevant html markup  and committed.

src/components/site/SubpageTabs.md
```diff
@@ -17,7 +17,23 @@ import Section from "../basic/containers/Section";
...
+        class="dg_subpage-tabs-btn"
+        type="button"
+        id="dg_menu-trigger-btn"
+        aria-expanded="true"
+      >
+        <h2 class="dg_accordion_buttontext-holder">In This Section</h2>
+      </button>
+      <ul class="dg_tab-items">
+        <li class="active"><a href="/home">Newsroom</a></li>
+        <li class=""><a href="/home">Social Media Gallery</a></li>
```
martypowell: you shouldn't need these clases here, or maybe you want to do something like `dg_tab-items_item` so that you could style each li

theilman7: Removed the empty class and committed. I have a question regarding this item though. In my SubpageTabItem.js file, I have it setup as follows:
```
    <li className={TabItemClasses}>
      <a href={link}>{text}</a>
    </li>
```
Is there a way in the react file to setup if there is no class entered with TabItemClasses, it won't render an empty class="", or is there not really a way around that?

martypowell: @theilman7 looks like this is the way to do it, https://stackoverflow.com/a/48816255/1143670

But i wouldn't worry about it too much, I realized that i haven't really been doing my suggestion.

theilman7:
Regarding the question about the javascript, correct. This utilizes vanilla js.

theilman7:
@martypowell all previous comments were addressed. Ready for another look when you have a moment.

martypowell:
@theilman7 looks like there are a couple changes

- branch is out of date
- per our discussion with @sdurphy yesterday, this is how the component should look when collapsed

image

I believe the padding is the default-padding.

theilman7:
Brought branch up-to-date.

theilman7:
Updated correction to list item case in recent commit.

theilman7:
Updated styles so hover extends full width.

theilman7:
Updated styles so the text color no longer blends with background on hover.

src/components/site/SubPageTabItem.js
```diff
@@ -0,0 +1,22 @@
+import React from "react";
+import PropTypes from "prop-types";
+import classnames from "classnames";
+
+const SubPageTabItem = props => {
+  const { text, link, className } = props;
+  const TabItemClasses = classnames(className);
```
martypowell: local variables are usually lower case. I personally like the name `cssClasses` in this case. but you can probably come up with something better.

src/components/site/SubPageTabs.js
```diff
@@ -0,0 +1,34 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+const SubPageTabs = props => {
+  const { text, id, isExpanded = true } = props;
+  return (
+    <div className="dg_subpage-tabs-container">
+      <div className="container">
```
martypowell: I think we were going to try and use sass with @extend to try and accomplish this, and to remove this particular div?

src/components/site/SubpageTabs.md
```diff
@@ -0,0 +1,39 @@
...
+```jsx
+import SubPageTabs from "./SubPageTabs";
+import SubPageTabItem from "./SubPageTabItem";
+import Section from "../basic/containers/Section";
+
+<Section>
+  <SubPageTabs text="In This Section" id="dg_menu-trigger-btn">
+    <SubPageTabItem className="active" text="Newsroom" link="/home" />
+    <SubPageTabItem text="Social Media Gallery" link="/home" />
+    <SubPageTabItem text="Media Contact Information" link="/home" />
```
martypowell: you probably should make some fake links here so the reflect something similar to the text

martypowell (changes requested):
please see comments

src/components/site/SubpageTabs.md
theilman7: Updated and committed.

src/components/site/SubPageTabItem.js
theilman7: This is in reference to `TabItemClasses`, correct? If so, would `itemClasses` work/ make sense?

martypowell: works for me, i like putting in css in the name, just because maybe you have something like vehicleClasses which relates to something different, but this is readable enough in my opinion.

theilman7: Sounds good. Updated and pushed up in the latest commit.

src/components/site/SubPageTabs.js
theilman7: Updated to remove the `container` class. Set `dg_subpage-tabs `to handle the container width at the proper breakpoint and added the proper padding and margins to match.

theilman7:
Updated branch as it was out of date.

martypowell (changes requested):
One minor change, `.dg_subpage-tabs-container` should have some padding too (at least on smaller screens)

image

theilman7:
@martypowell Wouldn't let me respond directly to your comment above, but added a `margin` of 15px left and right on mobile. 15px will keep the borders in line with the 15px padding used in the `container` class across the board.

theilman7:
@martypowell  Ready for another look.

martypowell:
> @martypowell Ready for another look.

looking now.

src/styles/partials/components/_subpage-tabs.scss
```diff
@@ -0,0 +1,156 @@
...
+    padding-left: 0px;
+
+    li {
+      float: left;
+
+      a {
+        display: block;
+        color: $white;
+        text-align: center;
+        padding: $padding-large 16px;
```
martypowell: should this be default padding?

src/styles/partials/components/_subpage-tabs.scss
```diff
@@ -0,0 +1,156 @@
...
+    max-width: $max-container-width;
+  }
+}
+
+@media screen and (max-width: $large-breakpoint) {
+  .dg_subpage-tabs {
+    border-top: 1px $gray solid;
+    border-bottom: 1px $gray solid;
+    font-weight: normal;
+    margin: $default-margin 15px $default-margin 15px;
```
martypowell: should this use a variable?

src/styles/partials/components/_subpage-tabs.scss
```diff
@@ -0,0 +1,156 @@
...
+
+@media screen and (min-width: $extra-large-breakpoint) {
+  .dg_subpage-tabs {
+    max-width: $max-container-width;
+  }
+}
+
+@media screen and (max-width: $large-breakpoint) {
+  .dg_subpage-tabs {
+    border-top: 1px $gray solid;
```
martypowell: i believe 1px border widht is a variable

theilman7: Will take a look and update with the variable.

theilman7: Just looked at commits, didn't notice you had already updated and committed. Merged in to local branch.

src/styles/partials/components/_subpage-tabs.scss
theilman7: There is a variable for 15px padding, but not for 15px margin in the _spacing.scss file. Let me know if you would like me to add this or leave as-is.

src/styles/partials/components/_subpage-tabs.scss
theilman7: Updated to default padding variable.

theilman7: @martypowell Updated and committed.

PR #213, opened 2020-01-16
Allow Children in Link button

martypowell:
Allows for children to be passed under the text of an Icon Link. If not children are passed nothing is rendered. This does not impact the current implementation but allows for us to do some stuff for the services list.

No example is provided but here is a screen shot of a quickly mocked up service link.

image


tmccoy529 (approved):
Looks good

PR #214, opened 2020-01-22
Link button anchor

martypowell:
Added ability to make a secondary button a link element. This is used on the search results page.

### DIFF
--- a/src/components/basic/Button.js
+++ b/src/components/basic/Button.js
@@ -1,27 +1,43 @@
-import React from 'react';
-import PropTypes from 'prop-types';
-import classnames from 'classnames';
+import PropTypes from "prop-types";
+import React from "react";
+import classnames from "classnames";
 
-const Button = (props) => {
-	const { icon, className, type = 'button', text } = props;
-	const buttonCssClasses = classnames('dg_button', className);
-	const iconCssClasses = classnames('dg_button_icon', icon);
-	const buttonIcon = icon ? <i aria-hidden="true" className={iconCssClasses} /> : null;
-	return (
-		<button type={type} className={buttonCssClasses}>
-			{buttonIcon}
-			{text}
-		</button>
-	);
+const Button = props => {
+  const {
+    as: As = "button",
+    icon,
+    className,
+    type = "button",
+    text,
+    ...rest
+  } = props;
+  const buttonCssClasses = classnames("dg_button", className);
+  const iconCssClasses = classnames("dg_button_icon", icon);
+  const buttonIcon = icon ? (
+    <i aria-hidden="true" className={iconCssClasses} />
+  ) : null;
+  return As === "button" ? (
+    <As type={type} className={buttonCssClasses}>
+      {buttonIcon}
+      {text}
+    </As>
+  ) : (
+    <As className={buttonCssClasses} {...rest}>
+      {buttonIcon}
+      {text}
+    </As>
+  );
 };
 
 Button.propTypes = {
-	/** Button Text */
-	text: PropTypes.string.isRequired,
-	/** Font Awesome icon class */
-	icon: PropTypes.string,
-	/** Button Type - either `button`, `submit`, `reset` */
-	type: PropTypes.string
+  /** Button Text */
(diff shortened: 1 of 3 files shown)


src/components/basic/Button.js
```diff
@@ -1,27 +1,43 @@
...
-	/** Font Awesome icon class */
-	icon: PropTypes.string,
-	/** Button Type - either `button`, `submit`, `reset` */
-	type: PropTypes.string
+  /** Button Text */
+  text: PropTypes.string.isRequired,
+  /** Font Awesome icon class */
+  icon: PropTypes.string,
+  /** Button Type - either `button`, `submit`, `reset` */
+  type: PropTypes.string,
```
theilman7, line 38: Just thinking out loud, should this be set as required? I'm assuming this defaults to button if there is no `type` property assigned to it.

martypowell: since it can be a link which does not have a type, it can't be required (at least i dont think).

martypowell: this quesiton, probably is a good, should this be it's own component or not question.

martypowell (commented):
why does it do this?

src/styles/partials/components/_button.scss
```diff
@@ -1,108 +1,112 @@
...
+    background: $pale-blue;
+  }
+
+  &:disabled,
+  &.disabled {
+    border-color: $gray-light;
+  }
+}
+
+a.dg_button-secondary {
```
theilman7: Wondering if the `text-decroation: none;` property should be nested under `.dg_button, .dg_button-secondary`. Doing this way, it would apply to both button classes should either be tied to an a tag.

src/styles/partials/components/_button.scss
```diff
@@ -1,108 +1,112 @@
...
-		background: $pale-blue;
-	}
-
-	&:disabled,
-	&.disabled {
-		border-color: $gray-light;
-	}
+  &:hover,
+  &.hovered {
+    background: $vibrant-blue;
```
theilman7, line 5: Taking in to account we'll be using these classes for `a` tags as well, I noticed when setting an existing `button` to an `a` tag, it also inherits the `color` property, originally associated with the default tag. This makes the `color` blend in with the background. I would recommend setting a `color` property here to compensate.

src/styles/partials/components/_button.scss
martypowell: we could, this only applies to the link, do you think a comment would help out here?

src/styles/partials/components/_button.scss
martypowell: @theilman7 can you show me this one?

theilman7: @martypowell Certainly can. I'll meet with you tomorrow morning. Circling back, this only appears to affect the Button Primary component... or more specifically `.dg_button`.

src/styles/partials/components/_button.scss
theilman7: If this will only apply to a link using the `dg_button-secondary ` class now and in to the future, then it should be fine as-is without a comment added. If down the road, we would like to use `dg_button` class with a link as well, in that case I would nest.

martypowell: I think for now this is the only use case. Will add a comment.

src/components/basic/Button.js
martypowell: ended up removing text decoration from any state for a secondary button.

src/styles/partials/components/_button.scss
martypowell: ended up removing text decoration from any state for a secondary button.

src/styles/partials/components/_button.scss
theilman7: After discussing with Marty, this will be a non-issue since we do not plan on using `.dg_button` as a link button.

theilman7 (commented):
After discussing with Marty, this will be a non-issue since we do not plan on using .dg_button as a link button.

theilman7 (approved):
This looks good.

PR #216, opened 2020-01-23
Fixed casing for breadcrumb item

martypowell:
Casing in imports was causing Now to fail in this build.

These changes ensure they are consistent.

Update this change also removes better code hub from the build, since we aren't using it.


theilman7:
Taking a look now and will follow up with any feedback.

theilman7 (approved):
Looks good to me. Casing of component names appears to be fixed as the pull request name suggests.

PR #217, opened 2020-01-28
Style updates for collapse

tmccoy529:
Addresses Bug 15858 in TFS: Add Margin Between Filter and Search Area <768 pixels

### DIFF
--- a/src/components/basic/Collapse.md
+++ b/src/components/basic/Collapse.md
@@ -1,9 +1,28 @@
 ### Default (Expanded)
 
 ```jsx
+import Checkbox from "./form/Checkbox";
+
 <Collapse id="sample-collapse" header="Default Example">
-  <p>Some really compelling content goes here.</p>
-</Collapse>
+  <Checkbox
+    id="Checkbox1"
+    name="Checkbox1"
+    label="Show all popular services"
+    value="true"
+  />
+  <Checkbox
+    id="Checkbox2"
+    name="Checkbox2"
+    label="Show popular services for Rec and Park"
+    value="true"
+  />
+  <Checkbox
+    id="Checkbox3"
+    name="Checkbox3"
+    label="Show popular services for Corrections"
+    value="true"
+  />
+</Collapse>;
 ```
 
 Html Snippet:
--- a/src/styles/partials/components/_collapse.scss
+++ b/src/styles/partials/components/_collapse.scss
@@ -1,5 +1,6 @@
 .dg_collapse {
   border-top: $border-width-large solid $gray-light;
+  margin-bottom: $margin-larger;
 
   &:last-child {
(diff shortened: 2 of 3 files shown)


src/styles/partials/components/_collapse.scss
```diff
@@ -34,4 +34,8 @@
   .dg_accordion_buttontext-holder {
     margin-left: $margin-large;
   }
+
+  @media screen and (max-width: $medium-breakpoint) {
```
theilman7: @martypowell @tmccoy529 Does the spacing issue occur at tablet size as well/ do we want to cover that potential occurrence? If so, I would set the breakpoint to `$large-breakpoint` If not, this looks fine as-is.

theilman7:
Looks good to me, just have one question listed above.

src/styles/partials/components/_collapse.scss
tmccoy529: so actually it looks like we want it there at all times not on a break point. So this has been updated to reflect that instead.

src/components/basic/Collapse.md
```diff
@@ -1,9 +1,28 @@
...
+  <Checkbox
+    id="Checkbox3"
+    name="Checkbox3"
+    label="Show popular services for Corrections"
+    value="true"
+  />
+</Collapse>;
 ```
 
 Html Snippet:
```
martypowell, line 28: did the snippet change?

src/styles/partials/components/_collapse.scss
```diff
@@ -33,5 +34,6 @@
 
   .dg_accordion_buttontext-holder {
     margin-left: $margin-large;
+    margin-top: $margin-large;
```
martypowell: This rule isn't doing anything because this element displays is inline.

I think that's ok thought because this looks right to me as is. In this case I think the padding is doing ok in this instance?


image


Maybe there is something I'm missing?

src/components/basic/Collapse.md
tmccoy529: I removed the extra collapse as I didnt think it was needed but can add it back if you want to see the spacing again.

src/styles/partials/components/_collapse.scss
tmccoy529: No that was a mistake and has been removed

src/components/basic/Collapse.md
martypowell: got it.

martypowell (dismissed):
@tmccoy529 can you bump the version up one? Think it will be .80.

tmccoy529:
@martypowell version bumped to 80

martypowell:
merging on behalf of @tmccoy529 will be pushed up shortly.

PR #224, opened 2020-02-03
Scrollable Div Component

martypowell:
A disclaimer is shown https://beta.baltimorecountymd.gov/departments/budfin/taxpayerservices/taxsearch.html on this page. A disclaimer is a scroll-able div that allows a user to scroll in place to avoid taking up a ton of space.

This pr addresses this by adding styles and a component for this.

### DIFF
--- a/src/components/basic/containers/Scrollable.js
+++ b/src/components/basic/containers/Scrollable.js
@@ -1,0 +1,14 @@
+import React from "react";
+import classnames from "classnames";
+
+const Scrollable = props => {
+  const { children, className, ...rest } = props;
+  const cssClasses = classnames("dg_scrollable", className);
+  return (
+    <div className={cssClasses} {...rest}>
+      {children}
+    </div>
+  );
+};
+
+export default Scrollable;
--- a/src/components/basic/containers/Scrollable.md
+++ b/src/components/basic/containers/Scrollable.md
@@ -1,0 +1,37 @@
+```jsx
+<Scrollable>
+  <h3>User Terms and Conditions</h3>
+  <p>
+    Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod
+    tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam,
+    quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo
+    consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse
+    cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat
+    non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
+  </p>
+  <p>
+    Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod
+    tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam,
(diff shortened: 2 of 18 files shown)


theilman7:
Looking over this now. To confirm, the link attached is just to showcase the styles and not the actual scrolling function yet, correct?

src/components/basic/containers/Scrollable.js
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+import classnames from "classnames";
+
+const Scrollable = props => {
+  const { children, className, ...rest } = props;
+  const cssClasses = classnames("dg_scrollable", className);
```
theilman7: This is subjective, but wondering if something like `dg_scroll-container` or something along those lines would work. Scrollable seems like it could be vague to someone just coming in.

src/components/basic/containers/Scrollable.md
```diff
@@ -0,0 +1,37 @@
...
+  <p>
+    Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod
+    tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam,
+    quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo
+    consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse
+    cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat
+    non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
+  </p>
+</Scrollable>
+```
```
theilman7: I would add an HTML example to the .md file.

src/components/basic/containers/Scrollable.js
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+import classnames from "classnames";
+
+const Scrollable = props => {
+  const { children, className, ...rest } = props;
```
theilman7: What does the `...rest` variable do?

theilman7:
@martypowell Two very minor comments. Aside from those, looks good.

src/components/basic/containers/Scrollable.js
martypowell: @theilman7 haha, thats waht i orginally had, will update

src/components/basic/containers/Scrollable.md
martypowell: yep, will do.

src/components/basic/containers/Scrollable.js
martypowell: @theilman7 here is a pretty good explanation https://stackoverflow.com/a/43484565/1143670. 

In this case we are trying to make this container pretty flexible, so this allows us to add attributes to the container. 

The more you work with React, you will see this pattern used by others. It's not something I personally have a definitive article on (like this is the way you do it) but with practice has been something I personally have been doing.

sgrg1 (dismissed):
looks good

src/components/basic/containers/Scrollable.js
theilman7: Thanks for providing that information Marty!

theilman7:
Branch is out-of-date with integration. Once it has been updated I will approve. @martypowell

theilman7 (dismissed):
Approved.

PR #225, opened 2020-02-05
Adjsuted images per design with se kept in mind

martypowell:
During the demo we discovered taht the images still weren't working. That's because we only targeted floated images and some of the rules didn't apply because they were not in the main container. This addresses these issues based on feedback from design.

The nice 100% images we had, will require a little more work for the designers as they will need to ensure images are larger than there target containers to be 100% width

### DIFF
--- a/src/styles/partials/components/_images.scss
+++ b/src/styles/partials/components/_images.scss
@@ -1,4 +1,8 @@
-/* Handle images inside the main content section */
+/*
+    Handle images inside the main content section.
+    !important is used throughout the file to override the site executive feature of setting images width and height.
+    We want images to display to their natural size, and respond as screens become smaller.
+*/
 #dg_main-content {
   img,
   .dg_image {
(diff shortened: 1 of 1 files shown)


sgrg1 (dismissed):
looks good to me

tmccoy529 (dismissed):
We seem to be using !impotant a lot but Im assuming this is just unavoidable.

martypowell:
> We seem to be using !impotant a lot but Im assuming this is just unavoidable.

not idea, but in this case, a Site Executive user can speficy and width and height, and we need to override, this, that is the case for it in this case.

martypowell:
> > We seem to be using !impotant a lot but Im assuming this is just unavoidable.
> 
> not idea, but in this case, a Site Executive user can speficy and width and height, and we need to override, this, that is the case for it in this case.

saying that, i should probably add a comment.

tmccoy529 (approved):

PR #226, opened 2020-02-05
Bug#15869-Summary List Icons do not align with Headings

sgrg1:
This bug is dg_sidebar__section h3 has top margin 20px. That is why align was not correct . Override the margin-top on .dg_sidebar__section to fix this issue

### DIFF
--- a/src/styles/partials/components/_summary-list.scss
+++ b/src/styles/partials/components/_summary-list.scss
@@ -28,6 +28,15 @@
     justify-content: flex-start;
     align-items: flex-start;
     margin-bottom: $default-margin;
+    h1,
+    h2,
+    h3,
+    h4,
+    h5,
+    h6 {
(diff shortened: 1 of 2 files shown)


martypowell (changes requested):
So this works, but I think this is something we got wrong.  Instead of overriding by adding these new styles, we should set margin as follows in sidebar css on line 90.

margin: 0 0 $default-margin 0; this makes the bottom margin the default (which is the [one direction margin](https://csswizardry.com/2012/06/single-direction-margin-declarations/) we are looking for).

Then in the h2 right below the margin that we just changed you can set the top padding to the default padding to match xd.

martypowell (approved):

martypowell:
mergin on behalf of @sgrg1 since travis is still broken.

PR #228, opened 2020-02-06
Bug#15952-Global Nav List Item Icon Size and Position Incorrect

sgrg1:
Fix Font Aweson icon to 30pt.

### DIFF
--- a/src/styles/partials/components/_site-nav.scss
+++ b/src/styles/partials/components/_site-nav.scss
@@ -74,6 +74,7 @@ $mobile-menu-width: 245px;
           text-align: center;
           margin-right: $margin-small;
           width: 40px;
+          font-size: 30px;
         }
 
         &:after {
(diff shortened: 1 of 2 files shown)


src/styles/partials/components/_site-nav.scss
```diff
@@ -74,6 +74,7 @@ $mobile-menu-width: 245px;
           text-align: center;
           margin-right: $margin-small;
           width: 40px;
+          font-size: 30px;
```
martypowell: can you please create a variable for this in the _icon.scss variable file?

martypowell (changes requested):
@sgrg1 trying to verify with sam if the spacing has been fixed based on the icon size change, or wether we need to adjust the margin.

src/styles/partials/components/_site-nav.scss
sgrg1: done

martypowell (approved):

PR #230, opened 2020-02-06
Bug 15812

theilman7:
This references bug 15812. Removed the left margin on the` .dg_accordion `class. Also inferring the right should be removed as well, so the accordion extends the full frame and lines up evenly with content above and below it.

@martypowell @sdurphy

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@baltimorecounty/dotgov-components",
-  "version": "0.1.80",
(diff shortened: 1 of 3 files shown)


martypowell (dismissed):

martypowell:
@theilman7 can you just run a and `npm i` to resolve the confllict?

theilman7:
Yes, opening the branch now and will update in a second.

PR #233, opened 2020-02-12
Bug#15952

sgrg1:
According to Sam, the spacing between the global navigation menu list item icon and the text should be 20px. It is currently 15px. Added 5px in margin and that should give 20px

### DIFF
--- a/src/styles/partials/components/_site-nav.scss
+++ b/src/styles/partials/components/_site-nav.scss
@@ -72,7 +72,7 @@ $mobile-menu-width: 245px;
 
         i {
           text-align: center;
-          margin-right: $margin-small;
+          margin-right: $margin-small + $base-spacing-unit;
           width: 40px;
           font-size: $icon-site-nav-default-size;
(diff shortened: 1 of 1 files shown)


src/styles/partials/components/_site-nav.scss
```diff
@@ -72,7 +72,7 @@ $mobile-menu-width: 245px;
 
         i {
           text-align: center;
-          margin-right: $margin-small;
+          margin-right: $margin-small + $base-spacing-unit;
```
martypowell: @sdurphy is asking for the default margin here, which is set in the variable `$default-margin`. Please use that variable instead.

martypowell (changes requested):
minor change.

src/styles/partials/components/_site-nav.scss
sgrg1: used default-margin

martypowell (approved):

PR #246, opened 2020-03-10
Bug 16471 Newsroom

theilman7:
This PR addresses Bug 16471. There are two fixes here.

1. Adds `display: inline-block;` to .`dg_button` while it is within `.dg_card__footer`. This allows the button to achieve even spacing within its container.

2. I added `margin-bottom: 20px;` to `i` and `img` tags while children of the class `.dg_card__footer`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@baltimorecounty/dotgov-components",
-  "version": "0.1.100",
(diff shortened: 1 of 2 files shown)


martypowell (approved):

martypowell:
mergin on behalf of @theilman7

PR #250, opened 2020-03-12
Bug 16474 - Newsroom - News Cards (on Home Page) IE Issues

theilman7:
This PR addresses bug 16474.

1. Added `max-height: 300px;` and `height: 100%; `to `.dg_date-news-card`. This fixes a bug where IE doesn't calculate height in flex if there isn't a definitive height set.

2. Testing in browser stack revealed there is no longer an issue with the heading text being in all caps. This must have been inadvertently resolved with a previous PR.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@baltimorecounty/dotgov-components",
-  "version": "0.1.103",
+  "version": "0.1.104",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@baltimorecounty/dotgov-components",
-  "version": "0.1.103",
+  "version": "0.1.104",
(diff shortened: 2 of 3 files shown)


martypowell:
@theilman7 did we confirm this is responsive as well?

theilman7:
@martypowell Checking this now. Noticing the fix in ie is causing issues elsewhere when viewed on mobile in Chrome. Will be taking another look.

martypowell:
> @martypowell Checking this now. Noticing the fix in ie is causing issues elsewhere when viewed on mobile in Chrome. Will be taking another look.

sounds good @theilman7 i know at the time maybe it was okay enough because ie stinks, so it might be worth mentioning this to sam and see what we think.

theilman7:
@martypowell Once the issue with SE is cleared up and we can test the fix in the other repo, we should be able to cancel this PR. You can disregard this item for now.

theilman7:
Cancelling pull request.

PR #256, opened 2020-04-02
Bug 2030

theilman7:
This adds margin utilities to be used in certain scenarios throughout the site.


martypowell:
@theilman7 can we just do bottom for now? also looks like there might be some formatting issues.

theilman7:
@martypowell Rolling this back to only utilize bottom. Will commit shortly.

theilman7:
Removed all other spacing utility classes except for margin-bottom. Let me know if this is fine or if I should restore the other margin-bottom utilities as well. @martypowell

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@baltimorecounty/dotgov-components",
-  "version": "1.0.0",
+  "version": "1.0.01",
```
martypowell: 1.0.1

martypowell (changes requested):
cahnge version and we are good

martypowell (approved):

PR #258, opened 2020-04-14
Fixed Bottom Margin on Form "Groups"

martypowell:
We noticed some form elements were scrunched up while testing the checkbox styles.

image

This pr addresses this by targeting a consistent class available around each form field. It also adds a bottom margin of the element if it's the last child.


tmccoy529 (approved):

theilman7 (approved):
Looks good.

PR #261, opened 2020-04-29
User story 16939

theilman7:
Addresses margin issues of `.dg_alt-list`, when added as a child to `.dg_summary-list`.

### DIFF
--- a/src/styles/partials/components/_alt-list.scss
+++ b/src/styles/partials/components/_alt-list.scss
@@ -70,11 +70,12 @@ $icon-arrow-width: 32px;
 
 .dg_sidebar__section {
   .dg_alt-list {
+    margin: 0px;
     li {
       margin: $base-spacing-unit 0 0 0;
     }
     a {
-      margin: 0 0 0 0;
+      margin: 0px;
     }
   }
 }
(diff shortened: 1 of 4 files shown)


src/styles/partials/components/_alt-list.scss
```diff
@@ -68,6 +68,18 @@ $icon-arrow-width: 32px;
   }
 }
 
+.dg_sidebar__section {
```
martypowell: @theilman7 i thought we were able to make these changes regardless of it being inside of the sidebar?

theilman7: @martypowell So the component works fine in its current state within the footer, maintaining the margins etc. we want. The problem comes when it is added to the sidebar, due to a reset string. The `<li>` and `<a>` tags are both selected in the reset string with `.dg_sidebar__section li` and `.dg_sidebar__section a`. This kind of "double dips" by adding `margin: 0 0 20px 0;` to both the `li `and `a `tags. The simplest methods I could think of to work around the reset are as followed. I could be missing something though, so let me know if there is a more efficient method to accomplish this.

1. Removing the `li `and `a `tag from the reset, which I think could have ripple effects, if there is something this is resetting that it should.

2. Adding a class to all the 'li' tags within the alt-list component. This allows the elements to be selected more accurately, but would add additional markup.

3. Selecting the elements within the sidebar component, as shown here. Thought this would be the most non-invasive method.

martypowell:
this is being closed in favor of #263

PR #267, opened 2020-05-06
Fix Sidebar Containers

theilman7:
Creating PR to begin discussion on the sidebar. I added two content containers for content in the sidebar. One is a standard container, where items can just be placed in, the other is an option for a gray background. Let me know if this is heading int he right direction.

### DIFF
--- a/src/styles/partials/components/_summary-list.scss
+++ b/src/styles/partials/components/_summary-list.scss
@@ -4,12 +4,8 @@
       display: block;
       margin-bottom: $margin-larger;
 
-      i,
-      & > div {
-        width: 100%;
-      }
-
       i {
+        display: block;
         margin-bottom: $default-margin;
         text-align: left;
       }
@@ -32,10 +28,6 @@
     p {
       color: $gray;
     }
-
-    & > div {
-      width: 75%;
-    }
   }
 
   h3 {
@@ -43,10 +35,9 @@
   }
 
   i {
-    display: block;
     text-align: center;
     font-size: $h1-font-size;
-    width: 25%;
+    margin-right: $default-padding;
   }
 
   a {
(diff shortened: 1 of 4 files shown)


src/styles/partials/template/_sidebar.scss
```diff
@@ -18,7 +18,15 @@
   p {
     font-family: $default-heading-font;
     font-size: $smallest-heading-font-size;
-    margin: 0 0 $default-margin 0;
+  }
+
+  .dg_sidebar__content {
```
martypowell, line 23: @theilman7 have you adjusted this markup anywhere?

theilman7: @martypowell Not yet, just with inspect. I can copy this over and insert it on to the components page to test. Will follow up when that has been added shortly.

theilman7: @martypowell Pushed up some corrections and staged on the components page in beta.

src/styles/partials/components/_summary-list.scss
```diff
@@ -32,21 +28,16 @@
...
   h3 {
     color: $black;
   }
 
   i {
-    display: block;
     text-align: center;
     font-size: $h1-font-size;
-    width: 25%;
+    margin-right: $default-padding;
```
martypowell, line 40: this is  agood example why we should have one value for spacing margn / padding. which i think should have been 15px. That would be a big refactor now, so we can leave this as is.

src/styles/partials/template/_sidebar.scss
```diff
@@ -20,6 +20,15 @@
     font-size: $smallest-heading-font-size;
   }
 
+  .dg_sidebar__content {
+    padding: $padding-small $padding-large;
+  }
+
+  .dg_sidebar__content_gray {
```
martypowell: how does this differ from the rule above this? I think maybe we could name this a little better to describe it.

theilman7: Essentially all it does is add the option for a gray background. Maybe .dg_sidebar___content_gray_background? A bit long though.

martypowell (approved):

PR #269, opened 2020-05-20
Alt list arrow update

theilman7:
Minor update, changing the arrow styling for alt-list. Switching from f054 to f0a9.

### DIFF
--- a/src/styles/partials/components/_alt-list.scss
+++ b/src/styles/partials/components/_alt-list.scss
@@ -40,7 +40,7 @@ $icon-arrow-width: 32px;
     &:after {
       color: $gray-light;
       @include default-transition;
-      @include icon-content($icon-arrow-right);
+      @include icon-content($icon-arrow-right-circle);
       font-size: $icon-arrow-default-size;
       margin-left: $default-margin;
       position: absolute;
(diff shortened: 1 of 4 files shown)


src/styles/partials/components/_alt-list.scss
```diff
@@ -40,7 +40,7 @@ $icon-arrow-width: 32px;
     &:after {
       color: $gray-light;
       @include default-transition;
-      @include icon-content($icon-arrow-right);
+      @include icon-content($icon-arrow-right-circle);
```
martypowell: seems like you want to update the icon-arrow-right, instead of circle so it's update throughout?

martypowell: or is this only targeting links?

theilman7: @martypowell I don't think we want to do that, as icon-arrow-right variable is used for accordions currently. This just adds an updated icon to the alt-list component.

martypowell: to make it easier on the developers, can we rename this osmething maybe like `$icon-arrow-link` or something along those lines, something that might make a little more sense?

theilman7: Makes sense to me. Updated and committed.

theilman7: @martypowell Ready for a second look when you have a moment.

martypowell (approved):

PR #270, opened 2020-05-21
Status Banner Color Updates

theilman7:
There are a few things that will need to change, but wanted to get this in to PR for initial feedback. Let me know how you would like the colors formatted in relation to the project. In addition, let me know if the markup I have makes sense. Thinking I could possibly encapsulate the elements I selected by using `*`... but let me know your thoughts.

### DIFF
--- a/src/styles/partials/components/_alert.scss
+++ b/src/styles/partials/components/_alert.scss
@@ -42,7 +42,7 @@
   }
 
   &.emergency {
-    background: $error-color;
+    background: $emergency-color;
 
     a:hover {
       background: $white;
@@ -79,16 +79,30 @@
       background: $info-color;
       outline-color: $info-color;
 
+      .dg_alert__status {
+        color: $black;
+      }
+
+      .dg_alert__icon {
+        color: $black;
+      }
+
+      p {
+        color: $black;
+      }
+      * {
+        color: $black;
+      }
       a:hover {
-        background: $white;
+        background: $black;
         color: $info-color;
       }
     }
 
     &.warning {
-      background: $warning;
+      background: $warning-color;
       border-color: $black;
-      outline-color: $warning;
+      outline-color: $warning-color;
 
       * {
         color: $black;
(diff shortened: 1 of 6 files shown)


src/styles/partials/components/_alert.scss
```diff
@@ -79,16 +82,30 @@
...
+        color: $black;
+      }
+
+      .dg_alert__icon {
+        color: $black;
+      }
+
+      p {
+        color: $black;
+      }
```
martypowell: i dont think you need nay of these lines if you use *

src/styles/partials/components/_alert.scss
```diff
@@ -104,8 +121,22 @@
...
       outline-color: $error-color;
 
+      .dg_alert__status {
+        color: $black;
+      }
+
+      .dg_alert__icon {
+        color: $black;
+      }
+      * {
```
martypowell, line 115: think you can just use * here

src/styles/variables/_colors.scss
```diff
@@ -38,8 +38,11 @@ $gray-dark: mix(black, $gray, $gray-light-ratio);
 $gray-darkest: mix(black, $gray, $gray-lightest-ratio);
 $black: black;
 $white: white;
-
+$alice-blue: #E8F4FD;
```
martypowell: i'd go with $status-blue or something like that.

martypowell: i actually have mixed feelins about this, but i think htat naming convetion lets us know more about what these colors are, and to not really use them outside of their purpose.

martypowell (changes requested):
a couple minor suggestions.

src/styles/variables/_colors.scss
theilman7: Updated. Ready for another look @martypowell

martypowell: youll want to hit hit the other status colors too.

src/styles/partials/components/_alert.scss
theilman7: Did a sweep and cleaned up properties a bit. Take a look and let me know if this looks better. This should address your comment above as well. @martypowell

src/styles/variables/_colors.scss
theilman7: Did a big cleanup. Pushing up. @martypowell

martypowell (approved):

PR #276, opened 2020-06-30
Parks detail bug

theilman7:
Adds the `column-width` property for proper scaling on tablet and mobile. I can provide a link to the sample page off thread if needed.


martypowell (approved):
I'm not sure about the rems but ill defer to your judgement?

PR #277, opened 2020-07-01
Added overrides for phone directory

martypowell:
Title pretty much says it all.

Whatever couldn't be modified in the inclusion was manually overrided with css.

### DIFF
--- a/src/dotgov.scss
+++ b/src/dotgov.scss
@@ -93,10 +93,11 @@
 @import "./styles/overrides/ie-hacks";
 @import "./styles/overrides/se-form";
 @import "./styles/overrides/google-search-results";
+@import "./styles/overrides/phone-directory";
 
 /** experimental */
 @import "./styles/experimental/grid";
 
 * {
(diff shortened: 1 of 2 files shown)


src/dotgov.scss
```diff
@@ -93,10 +93,11 @@
 @import "./styles/overrides/ie-hacks";
 @import "./styles/overrides/se-form";
 @import "./styles/overrides/google-search-results";
+@import "./styles/overrides/phone-directory";
 
 /** experimental */
 @import "./styles/experimental/grid";
 
 * {
```
theilman7, line 101: What is this function doing?

martypowell: that imports the phone directory css into our main stylesheet...it's probably not ideal, but the quickest way to integrate right now. you guys could start making page specific files.

martypowell: @theilman7

theilman7:
Looks good.

PR #279, opened 2020-07-06
Social media buttons

tmccoy529:
This is adding the twitter and facebook social media buttons both branded in company colors and Baltimore county colors. These will not work without the dotgov-javascript-site.min.js file. Also this branch has been merged with the events branch. That is why its so large. Once the events have been merged into master this will shrink so it should probably wait until the events branch is done.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1341,6 +1341,11 @@
         }
       }
     },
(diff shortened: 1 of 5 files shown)


martypowell (dismissed):

sgrg1 (approved):
looks good

PR #280, opened 2020-07-07
Bug 17884

theilman7:
This fix addresses the uneven headings in the sidebar of pages, caused by inconsistent sizes of FontAwesome icons. Piggy-backed off the FontAwesome fa-fw class to address the issue, by adding a width and min-width to the i tag in the sidebar. The fix has been tested across two pages, and seems to work.

### DIFF
--- a/src/styles/partials/components/_summary-list.scss
+++ b/src/styles/partials/components/_summary-list.scss
@@ -35,10 +35,11 @@
   }
 
   i {
-    text-align: left;
+    text-align: center;
     font-size: $h1-font-size;
     margin-right: $default-padding;
-    width: 40px;
+    width: 45px;
+    max-width: 45px;
   }
 
   a {
--- a/src/styles/partials/components/_summary-list.scss
+++ b/src/styles/partials/components/_summary-list.scss
@@ -35,9 +35,10 @@
   }
 
   i {
-    text-align: center;
+    text-align: left;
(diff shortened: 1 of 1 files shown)


src/styles/partials/components/_summary-list.scss
```diff
@@ -38,6 +38,8 @@
     text-align: center;
     font-size: $h1-font-size;
     margin-right: $default-padding;
+    width: 45px;
+    max-width: 45px;
   }
```
tmccoy529, line 43: would these be $padding-largest?

theilman7: Wanted to discuss this one before making the update. The positive to this, it would be it makes the update pretty simple by using a variable that can be updated from one location in this instance. One negative I can see to updating a width property with the padding variable, is if there is ever a change to padding-largest, we would need to go back and update the unit for the width. That might not be bad in one instance, but if it opens the door to using spacing variables in other width instances, across multiple files, that could become tedious down the road. Thoughts?

tmccoy529: @theilman7 maybe we need width variables? Which might be biting off more than we can chew at the moment. So yeah I think leave it as is and we can revisit width variables later.

tmccoy529 (approved):

PR #293, opened 2020-07-29
Update google search

srutherfordbcg:
To support the google autocomplete search bar:
1) Update the input field to the div google CSE is looking for
2) Update dotgov-site.min.js to use beta version on demo page
3) Include jquery on demo page

### DIFF
--- a/styleguide.config.js
+++ b/styleguide.config.js
@@ -13,17 +13,17 @@ module.exports = {
       sections: [
         {
           name: "Colors",
-          content: "src/components/foundation/Colors.md"
+          content: "src/components/foundation/Colors.md",
         },
         {
           name: "Grid",
-          content: "src/components/foundation/Grid.md"
+          content: "src/components/foundation/Grid.md",
         },
         {
           name: "Typography",
-          content: "src/components/foundation/Typography.md"
-        }
-      ]
+          content: "src/components/foundation/Typography.md",
+        },
+      ],
     },
     {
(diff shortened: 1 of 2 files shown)


styleguide.config.js
```diff
@@ -13,17 +13,17 @@ module.exports = {
...
+          content: "src/components/foundation/Grid.md",
         },
         {
           name: "Typography",
-          content: "src/components/foundation/Typography.md"
-        }
-      ]
+          content: "src/components/foundation/Typography.md",
+        },
+      ],
```
ascire, line 26: Do we have a style guide that suggests we put commas on the last element in an object/array?

I get the rationale; I'm just curious if we have this documented.

ascire (approved):

styleguide.config.js
tmccoy529: Documented? Hahaha thats funny, and sadly no. Honestly Marty was using prettier as our way of kind of forcing these kinds of things which we were all fine with since it auto formats for us and as we came across a new thing we added a rule. Might be a way to dump those rules to a readme but I have looked.

srutherfordbcg: Yeah I didn't intentionally add these commas, but I was told to install the Prettier addon to my VS Code which automatically added these when I saved the file. There is a setting to control trailing commas - maybe we can take a dev meeting one day and review the settings everyone should be using in Prettier.

PR #299, opened 2020-08-12
Share padding

tmccoy529:
FIxed accordion spill over. Fixed share button padding.

### DIFF
--- a/src/styles/overrides/_se-form.scss
+++ b/src/styles/overrides/_se-form.scss
@@ -1,104 +1,109 @@
 .seform {
-    form {
+  form {
+    .seformRow {
+      margin-bottom: $default-margin;
+    }
 
-        .seformRow {
-            margin-bottom: $default-margin;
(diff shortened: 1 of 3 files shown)


src/styles/overrides/_se-form.scss
```diff
@@ -1,104 +1,109 @@
...
+.SEAFWrapper > .seText > div > span {
+  display: block;
+  margin-bottom: $default-margin;
+}
+
+@media screen and (max-width: $medium-breakpoint) {
+  .buttonSet {
+    padding-bottom: $padding-largest;
+  }
+}
```
tmccoy529, line 109: Looks like prettier did its reformatting magic.

tmccoy529: .buttonSet is the only thing that actually was added to this file.

sgrg1 (approved):

PR #305, opened 2020-08-21
Z-Index Update on Menu

theilman7:
This appears to fix the issue where the menu transitions in/ out of view when reducing the screen size past 768 px. Please feel free to test in IE to confirm this fix. To test, simply inspect element, remove the z-index property and reduce the screen size.


tmccoy529 (dismissed):

theilman7:
Minor update. Removed the default transition property. Seems to keep the size transition from occurring when reducing screen size.

tmccoy529 (approved):
looks good in ie and chrome

PR #308, opened 2020-08-27
removed hidden prop from a required field

tmccoy529:
This hidden property was preventing ie from displaying this red *. Im not honestly sure why we did it this way. It appears we hide the icon until we include the icon then we unhide it.....so I deleted that. Not sure what the point of that was but maybe Im missing something. Thankfully we have code reviews.


theilman7 (approved):
Looks good to me.

PR #309, opened 2020-09-01
Fancy box override

sgrg1:
added fancy box scsss file


src/styles/overrides/_fancybox.scss
```diff
@@ -0,0 +1,25 @@
...
+  display: inline-block;
+  margin-right: -100%;
+  padding: 2px 20px;
+  background: transparent; /* Fallback for web browsers that doesn't support RGBa */
+  background: rgba(0, 0, 0, 0.8);
+  -webkit-border-radius: 15px;
+  -moz-border-radius: 15px;
+  border-radius: 15px;
+  text-shadow: 0 1px 2px #222;
+  color: #fff;
```
theilman7, line 21: This should utilize the color variable `$white`.

src/styles/overrides/_fancybox.scss
```diff
@@ -0,0 +1,25 @@
...
+  font-weight: bold;
```
theilman7, line 22: Think this should translate over to the variable `$font-bold`.

src/styles/overrides/_fancybox.scss
```diff
@@ -0,0 +1,25 @@
...
+.fancybox-title-float-wrap .child {
+  display: inline-block;
+  margin-right: -100%;
+  padding: 2px 20px;
+  background: transparent; /* Fallback for web browsers that doesn't support RGBa */
+  background: rgba(0, 0, 0, 0.8);
+  -webkit-border-radius: 15px;
+  -moz-border-radius: 15px;
+  border-radius: 15px;
+  text-shadow: 0 1px 2px #222;
```
theilman7, line 20: Wondering if the color translates over to the variable `$gray-dark` in our colors library. @tmccoy529 thoughts?

src/styles/overrides/_fancybox.scss
```diff
@@ -0,0 +1,25 @@
...
+  margin-bottom: -35px;
+  z-index: 8050;
+  text-align: center;
+  width: 100%;
+}
+
+.fancybox-title-float-wrap .child {
+  display: inline-block;
+  margin-right: -100%;
+  padding: 2px 20px;
```
theilman7, line 14: I don't think we have a variable for 2px so that should be fine as-is, but for 20 px, you can use `$padding-large`.

src/styles/overrides/_fancybox.scss
```diff
@@ -0,0 +1,25 @@
+.fancybox-title-float-wrap {
+  position: absolute;
+  bottom: 0;
+  right: 50%;
+  margin-bottom: -35px;
+  z-index: 8050;
```
theilman7, line 6: @tmccoy529 I'm testing the styles as they are on prod (if they exist in some way in beta let me know) and not really seeing the z-index add any meaningful functionality. When removed, the text doesn't seem to disappear behind anything. Are you getting similar results? I'd test again on beta, but if the results are the same, this line can probably be removed.

tmccoy529:
Upon further review, these changes didnt seem to be needed after updating the fancybox include to its more recent version. This PR is no longer needed

PR #312, opened 2020-09-18
Update to Gold and Purple Header Custom Colors

theilman7:
Headline says it all. Changed the custom header color string to use `gold-light` and purple to use `purple-light`.


tmccoy529 (approved):
Looks good

PR #313, opened 2020-09-22
16137 button documentation

theilman7:
- Condensed button markup examples from the .md files. This involved moving items from IconButton and IconLink to the Button section.
- Removed the IconButton and IconLink .md files. 
- Separated markup examples between `button `and `a `tag use cases to make it more evident to users browsing the components library.

When looking through, let me know if anything is unclear or if you would like to see any supporting copy added. The goal here is to make it clear to users what markup to use and in which case.


tmccoy529 (commented):
Can you just bump the version up on this? After that its good to go.

theilman7:
Sounds good. I'll bump the version and push up the branch.

theilman7:
Just finished updating the version and pushing up the branch.

tmccoy529 (approved):

PR #318, opened 2020-10-23
Blog Pagination Styles

theilman7:
Added styles for the SE pagination module. Extends button styles to the SE classes and sets up the container to handle flex for easy alignment.


tmccoy529 (dismissed):

tmccoy529 (approved):

theilman7:
Added pagination styles.

PR #322, opened 2020-11-17
Add spacing to search icon

tmccoy529:


src/styles/partials/components/_table.scss
```diff
@@ -13,6 +13,13 @@
     padding: $padding-large;
   }
 
+  .td_weather_details_overflow {
+    max-width: 1000px; //This is excessively large because its exclusive to the weather closing section.
+    overflow: hidden;
+    white-space: nowrap;
+    text-overflow: ellipsis;
+  }
+
```
tmccoy529: This should probably be moved into a new file and out of table to be honest. Maybe a page specific file like _closings.scss. Thoughts?

srutherfordbcg: I agree, if this doesn't apply for all tables, but just weather/closing related, it would be better in a site specific file.

theilman7: Mulling this over, so I'll have to get back to you. On one hand, it would condense table styles into one section/ document. On the other, it could be the beginnings of placing all the weather styles together.

srutherfordbcg (dismissed):

src/styles/applications/_closings.scss
```diff
@@ -0,0 +1,6 @@
+.weather_details_overflow {
```
theilman7: What part of the weather page does this relate to? Checking it out on dev. @tmccoy529

src/styles/partials/components/_search.scss
```diff
@@ -17,9 +17,9 @@
 }
 
 .dg_search-input {
-    padding: $padding-small;
-    border: 2px solid #000;
-    font-size: 20px;
-    font-family: "Montserrat",sans-serif;
-    width: 100%;
+  padding: $padding-small;
+  border: 2px solid #000;
```
theilman7: Recommend using variables for the 2px and color.

src/styles/partials/components/_search.scss
```diff
@@ -17,9 +17,9 @@
...
+  font-size: 20px;
```
theilman7: Recommend using variable for 20px size if available.

src/styles/partials/components/_search.scss
```diff
@@ -17,9 +17,9 @@
...
+  font-family: "Montserrat", sans-serif;
```
theilman7: Check for variables associated with the sans-serif font family.

src/styles/partials/components/_search.scss
tmccoy529: done

src/styles/partials/components/_search.scss
tmccoy529: done

sgrg1 (approved):

PR #338, opened 2021-01-12
Class name

tmccoy529:


tmccoy529:
This allows the table elements to accept classes which was not allowed prior. Also updated the responsive table to reflect the changes made to the javascript. This no longer looks for a table ID but instead a table class. This allows us to have multiple tables on a page that will be responsive.

sgrg1 (approved):

PR #349, opened 2021-02-22
Bug15224 link underline browser differences

sgrg1:
implemented class for <a> tag so that we can only use class with <a> tag where we need a underline.


tmccoy529:
We have decided to scrap this bug fix

PR #359, opened 2021-04-15
Active list css changes

tmccoy529:
Added a new style for alternate list items to reflect current selected item


tmccoy529:
This was merged with an incomplete branch. Need to clean up before we can approve

PR #402, opened 2022-06-21
Update Information Alert HTML Description

theilman7:
This changes `alerts.md` to display the correct class of `.information` instead of` .info`.


theilman7:
Bumping this change to bring it back to the top of the list.

tmccoy529 (approved):

PR #406, opened 2022-06-29
adds additional styles for input field

bbarbourbcmd:
Needed to add a few additional styles for the input and elements surrounding the input on this page.

Original css for reference:
.js-trails-filter {
  padding: 10px;
}

.search-container-trails > div {
  padding: 0 10px;
  width:100%;
}

@media screen and (max-width: 768px) {
  .search-container-trails > div {
    padding: 0 20px;
    width:100%;
  }
}

### DIFF
--- a/src/styles/partials/components/_walking-trails.scss
+++ b/src/styles/partials/components/_walking-trails.scss
@@ -253,7 +253,7 @@
     margin-left: $default-margin;
   }
 
-  & > div {
+  .search-container_inner {
     align-items: center;
     display: flex;
     justify-content: center;
@@ -280,7 +280,7 @@
     }
 
     .search-container-trails {
-      & > div {
+      .search-container_inner {
         padding: 0 $padding-large;
       }
     }
--- a/src/styles/partials/components/_walking-trails.scss
+++ b/src/styles/partials/components/_walking-trails.scss
@@ -314,6 +314,10 @@
       }
     }
   }
+
+  .parks-container .park .health-coalition {
+    text-align: left;
+  }
 }
 
 @media screen and (max-width: $small-breakpoint) {
--- a/src/styles/partials/components/_walking-trails.scss
+++ b/src/styles/partials/components/_walking-trails.scss
@@ -7,6 +7,11 @@
     &.map-legend-visible {
       background: $white;
       display: block;
+      height: 525px;
+    }
+
+    .esriLegendLayer {
+      font-size: .7em;
     }
   }
 
--- a/src/styles/partials/components/_walking-trails.scss
+++ b/src/styles/partials/components/_walking-trails.scss
@@ -294,3 +294,9 @@
     }
   }
 }
+
+@media screen and (max-width: $small-breakpoint) {
+  .legend-container {
+    display: none;
+  }
+}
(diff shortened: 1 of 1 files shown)


src/styles/partials/components/_walking-trails.scss
```diff
@@ -269,11 +273,19 @@
       margin: $default-margin;
       width: 100%;
     }
+
+    .search-container-trails {
+      & > div {
```
theilman7: Not sure this `& > div` nesting is necessary. The class `.search-container-trails` should be enough.

theilman7: To be more specific, think the properties should be nested under `.search-container-trails`.

bbarbourbcmd: I went ahead and added a class to the div, so it's not just blank markup. It seemed like a bit of a cleaner solution to just add a class rather than rework styles, some of which have already been minified and added to dev.

Previous styles:
.search-container-trails > div {
    align-items: center;
    display: flex;
    justify-content: center;
    padding: 0 10px;
    width: 100%;
}

@media screen and (max-width: 768px) {
    .search-container-trails > div  {width:100%; padding: 0 20px;}
}

Here's the latest commit:
https://github.com/baltimorecounty/dotgov-components/pull/406/commits/ef85c8a0d6d4301a747f003a855d408bafe80b12

bbarbourbcmd:
Added a few more style changes to the pagination/button hover styles, added the health coalition icons back, and changed the color of the disclaimer. Here's the original css for comparison:

.filter-walking-trails .paging button:not(:disabled),
.show-records button,
.search-container-trails button {
    cursor: pointer;
}

.filter-walking-trails .paging .paging-numbers button.active {
    color: #243e5d;
    font-weight: 900;
}

.filter-walking-trails .paging .paging-numbers button.active:hover {
    cursor: default;
}

.filter-walking-trails .paging .paging-numbers button {
    text-decoration: none;
}

.filter-walking-trails .paging .paging-numbers button:hover {
    text-decoration: underline;
}

.show-records button.active {
    display: none;
}

.parks-container .park .health-coalition {
    display: inline-block;
    margin: 0 auto 10px;
    text-align: right;
}

#dg_main-content .parks-container .park .health-coalition img {
    margin: 0;
}

.filter-walking-trails+.disclaimer {
    background: #F7F7F7;
}

src/styles/partials/components/_walking-trails.scss
```diff
@@ -269,11 +306,25 @@
...
+        padding: 0 $padding-large;
+      }
+    }
   }
 }
 
 @media (max-width: 600px) {
-  .js-trails-filter {
-    width: 100%;
+  .filter-walking-trails {
```
tmccoy529: Do we have a variable for the 600px? If not then ignore this comment

bbarbourbcmd: We don't have one for that number, but I was able to test if the next closest breakpoint we had a variable for would work, and everything still looks ok. So I think it's a bit cleaner now and still works.

tmccoy529 (approved):

theilman7:
Added one comment.

PR #408, opened 2022-07-25
Updates calc property to properly use scss variable

bbarbourbcmd:
I realized that even though the code compiled, I wasn't using the right syntax and the scss variable was not getting compiled. 

**Original css for quick reference:**
.dg_leadership-card {
   ...
   height: calc(100% - $margin-large);
   ...
}

### DIFF
--- a/src/styles/partials/components/_card-container-leadership.scss
+++ b/src/styles/partials/components/_card-container-leadership.scss
@@ -3,8 +3,8 @@
   border: $border-width-small solid $gray;
   display: flex;
   flex-direction: column;
-  height: calc(100% - $margin-large);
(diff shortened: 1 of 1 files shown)


theilman7 (commented):
Does the variable in the expression need the hash tag in front of `{$margin-large)`?

bbarbourbcmd:
Yeah, it needs the hashtag. It needs the brackets to work inside of a calc() value and needs the hashtag to be considered valid. It doesn't compile without it.

calc(100% - $margin-large) compiles, but the variable doesn't get converted to a value and calc(100% - {$margin-large}) is invalid and won't compile.

theilman7 (approved):
Approved.

PR #409, opened 2022-07-26
adds mobile style to park container

bbarbourbcmd:
Noticed some parks in the list were not wrapping properly on mobile, so I added some styles to force the park listings to stack on mobile instead of wrap.


bbarbourbcmd:
Nesting has been updated

theilman7 (approved):
Thanks for providing context. This is approved.

PR #410, opened 2022-08-10
22679 accepted materials

bbarbourbcmd:
Here are the original styles pulled from the head of the page for reference:

```
.fa-recycle{color:green;}
.fa-ban{color:red;}
.fa-trash-alt{color:#515477;}
.fa-truck{color:#666;}
#BACO_table .fa{margin-bottom:30px;}
.key{border-top:3px dotted #ccc;}
.sr-only {position: absolute;width: 1px;height: 1px;padding: 0;margin: -1px;overflow: hidden;clip: rect(0,0,0,0);border: 0;}
```

I moved the .sr-only class into _utilities.scss since I imagine we are/will be using this class in various parts of the site.

### DIFF
--- a/src/dotgov.scss
+++ b/src/dotgov.scss
@@ -40,6 +40,7 @@
 /** components */
 
 /*** header ***/
+@import "./styles/partials/components/accepted-materials";
 @import "./styles/partials/components/brand-text";
 @import "./styles/partials/components/site-status";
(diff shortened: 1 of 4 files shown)


theilman7 (approved):
Looks good to me. Think its ready to test in dev.

PR #411, opened 2022-08-16
Duplicate Resource Blog Image Utilities

theilman7:
This moves the image utilities (image_left and image_right) into the new template, to avoid breaking existing images. I also included a brief note to ensure we use the new classes in the future and sunset the old.


tmccoy529 (approved):

bbarbourbcmd:
Looks good from what I can tell

PR #415, opened 2022-10-13
Added ul to the alert property styles

theilman7:
This addresses Bug 22864. Nests the `<ul>` element so it inherits all the styling associated with the alert container.


bbarbourbcmd:
Code-wise this looks good to me. Not sure why Vercel failed to deploy, but I think all the code is syntactically correct, so I don't believe it's related to your work here.

tmccoy529 (approved):

PR #426, opened 2022-12-28
updates styles for the removal of Beta text in header of dotgov template

bbarbourbcmd:
I removed the right border on the wrapper that holds the "Baltimore county government" text and also adjusted the bottom margins of the header to remove extra space sans the Beta text.


theilman7 (approved):
Looks good! Feel free to deploy to test in dev, if you haven't already.

PR #428, opened 2023-02-24
Adds styles for executive news podcast feed

bbarbourbcmd:
Although it might seem a little like overkill now, I created a new component file for podcast styles in case the scope increases and Communications requests major styling updates with their new focus on podcasts.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@baltimorecounty/dotgov-components",
-  "version": "1.0.6",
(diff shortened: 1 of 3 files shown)


tmccoy529:
Hahaha I'm all about making new components especially if there is a chance we will expand. Better to over do then under do and than have to scramble later. Can you just increase the version by 1 and aside from that it looks good to me.

tmccoy529 (approved):

